PR #207, opened 2020-08-07
e2e refactor

skeeey:


skeeey:
see https://github.com/open-cluster-management/backlog/issues/4135

PR #215, opened 2020-08-11
fix image tag for release-2.1

ldpliu:


qiujian16:
i do not think you need this for now. We can change this when we freeze the code

PR #219, opened 2020-08-12
Add validating webhook for managedclusterclaim

elgnay:


qiujian16:
will need e2e for this, but i think we need installer to be updated by tomorrow.

/approve
/lgtm

PR #228, opened 2020-09-02
add clusterID label for managedCluster

zhiweiyin318:
task: https://github.com/open-cluster-management/backlog/issues/5021

### DIFF
--- a/pkg/klusterlet/clusterinfo/clusterinfo_controller.go
+++ b/pkg/klusterlet/clusterinfo/clusterinfo_controller.go
@@ -77,11 +77,11 @@ func (r *ClusterInfoReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 	// Get version
 	newStatus.Version = r.getVersion()
 
-	// Get distribution info
-	newStatus.DistributionInfo, err = r.getDistributionInfo()
+	// Get distribution info and ClusterID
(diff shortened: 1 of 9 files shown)


pkg/klusterlet/clusterinfo/clusterinfo_controller.go
```diff
@@ -531,24 +534,53 @@ func (r *ClusterInfoReconciler) getDistributionInfo() (clusterv1beta1.Distributi
...
-		// status.desired.version
-		DesiredVersion: desiredVersion,
-		// true when the first status.conditions element with type === Failing has status === True
-		UpgradeFailed: upgradeFailed,
-		// the value of the version property from each element in status.availableUpdates
-		AvailableUpdates: availableVersions,
+	return ocpDistributionInfo, nil
+}
+
+func (r *ClusterInfoReconciler) getOCPClusterID() (string, error) {
```
qiujian16, line 546: seems to me that getOCPDistributionInfo already gives us the clusterID. How about we also return cluster ID from getOCPDistributionInfo?

qiujian16 (commented):
I think  we need an e2e test for this

pkg/klusterlet/clusterinfo/clusterinfo_controller.go
zhiweiyin318: done. using  `getOCPDistributionInfo` can reduce one API call.

zhiweiyin318:
> I think we need an e2e test for this

done

PR #240, opened 2020-09-14
fix e2e fail using kind

zhiweiyin318:


qiujian16:
i have made some change on another PR. i think it should be fine

PR #245, opened 2020-09-15
add e2e for clusterset

ldpliu:


test/e2e/util/util.go
```diff
@@ -309,6 +317,29 @@ func SetStatusType(obj *unstructured.Unstructured, statusType string) error {
 
 	return nil
 }
+func SetSubjects(obj *unstructured.Unstructured, newSubject rbacv1.Subject) error {
```
qiujian16, line 320: one more line above

test/e2e/util/util.go
```diff
@@ -621,3 +652,6 @@ func podConditionsReady(pod corev1.Pod) error {
 
 	return nil
 }
+func GenerateClusterRoleBindingName(clusterName string) string {
```
qiujian16, line 655: one more line above, Is this func a generic func that also have been used in controller?

test/e2e/clusterset_test.go
```diff
@@ -0,0 +1,125 @@
...
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+		})
+		ginkgo.It("clusterrolebinding should be auto deleted successfully", func() {
+			gomega.Eventually(func() (interface{}, error) {
+				clusterroleBindingName := util.GenerateClusterRoleBindingName("cluster1")
+				return util.HasClusterResource(dynamicClient, clusterRoleBindingGVR, clusterroleBindingName)
+			}, eventuallyTimeout, eventuallyInterval).Should(gomega.BeFalse())
+		})
+		ginkgo.It("delete clusterrole/binding successfully", func() {
+			err = util.DeleteClusterResource(dynamicClient, clusterRoleGVR, clusterrole.GetName())
```
qiujian16, line 119: i am not sure the intent here. clean up?

test/e2e/util/util.go
ldpliu: changed use same func with controller.

test/e2e/clusterset_test.go
ldpliu: yes, clean up

qiujian16: how about using aftereach?

ldpliu: The clusterrole and clusterrolebinding is created in "ginkgo.It", and I think it's not need to create/update them in before each and aftereach.  It's just need to create once. what's your opinion?

qiujian16: I think having aftereach to clean all the test data seems reasonable. You could use label to delete all the test data. So that would also ease your  life when you have another case, such as "updating clusterset".

PR #247, opened 2020-09-23
Clusterset cluster

ldpliu:


pkg/controllers/clusterset/clustersetmapper/clustersetmapper.go
```diff
@@ -56,40 +58,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
-
-				klog.V(5).Infof("List managedClusterSet %+v", requests)
-				return requests
-			}),
-		})
-	if err != nil {
-		return nil
-	}
-
-	if err = c.Watch(&source.Kind{Type: &clusterv1alpha1.ManagedClusterSet{}},
```
qiujian16, line 92: I think you still need to watch managedClusterSet, what if a managedClusterSet is  added

pkg/helpers/clustersetmapper.go
```diff
@@ -43,45 +44,67 @@ func (c *ClusterSetMapper) DeleteClusterSet(clusterSetName string) {
...
+		}
+		clusters.Delete(clusterName)
+		if len(clusters) == 0 {
+			delete(c.clusterSetToClusters, clusterset)
+		}
+	}
+
+	return
 }
+func (c *ClusterSetMapper) UpdateClusterInClusterSet(clusterName, clusterSetName string) {
```
qiujian16, line 67: one space above, and let's doc how it works.
this func updates clusterset to cluster mapping. if a the clusterset of a cluster is changed, this func remove the previous mapping and  add a new one.

PR #271, opened 2020-12-09
check ocp311 version

zhiweiyin318:
issue: https://github.com/open-cluster-management/backlog/issues/6965

### DIFF
--- a/pkg/klusterlet/clusterinfo/clusterinfo_controller.go
+++ b/pkg/klusterlet/clusterinfo/clusterinfo_controller.go
@@ -16,6 +16,7 @@ import (
 	"github.com/prometheus/common/log"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


pkg/klusterlet/clusterinfo/clusterinfo_controller.go
```diff
@@ -471,6 +472,11 @@ func (r *ClusterInfoReconciler) getOCPDistributionInfo() (clusterv1beta1.OCPDist
 	ocpDistributionInfo := clusterv1beta1.OCPDistributionInfo{}
 	obj, err := r.ManagedClusterDynamicClient.Resource(ocpVersionGVR).Get(context.TODO(), "version", metav1.GetOptions{})
 	if err != nil {
```
qiujian16, line 474: should we have a more careful err check? And unit test please.
on vanilla kube, the version should not set. On ocp, it should set.

zhiweiyin318: `getOCPDistributionInfo` is called only when the cluster is OCP.  so I think `NotFound` is enough for check OCP 3.11 here.
UT is added.

PR #283, opened 2021-02-03
Fix inventory

qiujian16:
https://github.com/open-cluster-management/backlog/issues/8971


zhiweiyin318:
my understanding of the issue is that we also need to remove the `spec.ClusterDeployment` in `BareMetalAsset`. right? @jnpacker

jnpacker (approved):
lgtm

PR #286, opened 2021-02-05
detect OSD cluster

zhiweiyin318:
related issue:  https://github.com/open-cluster-management/backlog/issues/8841

### DIFF
--- a/pkg/klusterlet/clusterinfo/clusterinfo_controller.go
+++ b/pkg/klusterlet/clusterinfo/clusterinfo_controller.go
@@ -341,6 +341,8 @@ func (r *ClusterInfoReconciler) getVendor(gitVersion string, isOpenShift bool) (
 		return
 	case strings.Contains(gitVersion, string(clusterv1beta1.KubeVendorICP)):
 		kubeVendor = clusterv1beta1.KubeVendorICP
+	case r.isOpenshiftDedicated():
+		kubeVendor = clusterv1beta1.KubeVendorOSD
 	case isOpenShift:
 		kubeVendor = clusterv1beta1.KubeVendorOpenShift
 	default:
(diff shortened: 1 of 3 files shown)


pkg/klusterlet/clusterinfo/clusterinfo_controller.go
```diff
@@ -386,6 +390,32 @@ func (r *ClusterInfoReconciler) getCloudVendor() clusterv1beta1.CloudVendorType
...
+	Resource: "subjectpermissions",
+}
+
+const (
+	dedicatedAdmin    = "dedicated-admins"
+	rbacPermissionsNs = "openshift-rbac-permissions"
+)
+
+func (r *ClusterInfoReconciler) isOpenShiftDedicated() bool {
+	_, err := r.ManagedClusterDynamicClient.Resource(subjectpermissionsGVR).
```
qiujian16, line 407: could we just check the existence of this api resources? With this, you would need additional rbac

pkg/klusterlet/clusterinfo/clusterinfo_controller.go
```diff
@@ -342,7 +342,11 @@ func (r *ClusterInfoReconciler) getVendor(gitVersion string, isOpenShift bool) (
 	case strings.Contains(gitVersion, string(clusterv1beta1.KubeVendorICP)):
 		kubeVendor = clusterv1beta1.KubeVendorICP
 	case isOpenShift:
-		kubeVendor = clusterv1beta1.KubeVendorOpenShift
+		if r.isOpenShiftDedicated() {
```
qiujian16, line 345: this looks strange. why not using a case before this case?

qiujian16:
i think product in claim also needs to be handled.

pkg/klusterlet/clusterinfo/clusterinfo_controller.go
zhiweiyin318: yes. modified to use group name `managed.openshift.io` to detect if the cluster is osd, like using group name `project.openshift.io `to detect if the cluster is OCP.

pkg/klusterlet/clusterinfo/clusterinfo_controller.go
zhiweiyin318: done

zhiweiyin318:
> i think product in claim also needs to be handled.

done

pkg/klusterlet/clusterinfo/clusterinfo_controller.go
```diff
@@ -719,43 +743,53 @@ func (r *ClusterInfoReconciler) GetClusterClaims() ([]*clusterv1alpha1.ClusterCl
...
 	cloudVendor := r.getCloudVendor()
 	gitVersion := strings.ToUpper(kubeVersion)
+
+	// deal with product and platform
 	switch {
 	case strings.Contains(gitVersion, string(clusterv1beta1.KubeVendorIKS)):
-		// IKS
-		platform = string(clusterv1beta1.KubeVendorIKS)
+		// IBM Cloud Platform + IKS
+		platform = IBMPlatformType
```
qiujian16, line 753: I do not think we should do this in this release. UI maybe affected.

qiujian16 (commented):
/approve
/lgtm

PR #293, opened 2021-02-18
Do not apply owner/group for AppSub when from AppMgr

jnpacker:
* When the AppSub AppMgr controller creates a subscription we do not stamp a userID and userGroup on it.  This will be handled by the AppMgr controller.
* Need to replace `open-cluster-management` in the serviceAccount name with the `namespace` that the webhook is running in.

### DIFF
--- a/cmd/webhook/app/server.go
+++ b/cmd/webhook/app/server.go
@@ -45,6 +45,9 @@ func toAdmissionResponse(err error) *v1beta1.AdmissionResponse {
 // admitFunc is the type we use for all of our validators and mutators
 type admitFunc func(v1beta1.AdmissionReview) *v1beta1.AdmissionResponse
 
+// AppMgr service account, we do not overwrite the user, as the subscription will handle this.
+const UserIDforAppMgr = "system:serviceaccount:open-cluster-management-agent-addon:klusterlet-addon-appmgr"
+
 // serve handles the http portion of a request prior to handing to an admit
 // function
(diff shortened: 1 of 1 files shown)


jnpacker:
/cc @zhiweiyin318 @qiujian16 

FYI @rokej

cmd/webhook/app/server.go
```diff
@@ -108,6 +111,13 @@ func (a *admissionHandler) mutateResource(ar v1beta1.AdmissionReview) *v1beta1.A
 	}
 	annotations := crd.GetAnnotations()
 
+	// Do not change the userInfo if the object is being created by the AppMgr.
+	// This value is stamped onto the Subscription object by Appmgr (Only for subscriptions)
+	if crd.Kind == "subscription" && ar.Request.UserInfo.Username == UserIDforAppMgr {
```
qiujian16, line 116: Subscription?

@zhiweiyin318 as we discussed, the resource here should not be crd.

qiujian16 (commented):
it needs some tests also.

cmd/webhook/app/server.go
jnpacker: Is it somewhere else in the code then? As I'm looking to NOT block when the resource type is a "Subscription" kind and the userID is the AppMgr service account.

qiujian16: Yeah, i think the kind should be capitalized

jnpacker:
This code is no longer valid. Will rewrite and add some tests.

PR #310, opened 2021-03-15
upgrade hive

ldpliu:
https://github.com/open-cluster-management/backlog/issues/9948


zhiweiyin318:
do we also need to update the CRD related with Hive in deploy dir?

pkg/webhook/denynamespace/admission_test.go
```diff
@@ -6,10 +6,15 @@ import (
 	"github.com/stretchr/testify/assert"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
```
qiujian16, line 9: @zhiweiyin318 do we use this admission anywhere?  should we consider removing these codes?

PR #311, opened 2021-03-16
add clusterview rules to cluster admin and view roles

zhiweiyin318:
1. add clusterview rules into cluster admin/view clusterrole.
2. if user has  managedclustersets/join create role of a clusterset, add the user to the clusterset rolebinding to add cluster admin roles to the user.
3. user can only  get/list clusterview resources if user has get/list cluster roles.
4. add ut and fix e2e issue.


zhiweiyin318:
retest

qiujian16 (commented):
/lgtm

PR #312, opened 2021-03-17
detect IBM power and z arch managedCluster

zhiweiyin318:
show CloudLabel `IBM Power` for IBM power , `IBM Z` for IBM Z.

related issue: https://github.com/open-cluster-management/backlog/issues/8967

### DIFF
--- a/pkg/apis/internal.open-cluster-management.io/v1beta1/clusterinfo_types.go
+++ b/pkg/apis/internal.open-cluster-management.io/v1beta1/clusterinfo_types.go
@@ -161,6 +161,10 @@ type CloudVendorType string
 const (
 	// CloudVendorIBM IBM
 	CloudVendorIBM CloudVendorType = "IBM"
+	// CloudVendorIBMZ IBM s360x
+	CloudVendorIBMZ CloudVendorType = "IBMZPlatform"
+	// CloudVendorIBMP IBM Power
+	CloudVendorIBMP CloudVendorType = "IBMPowerPlatform"
 	// CloudVendorAWS Amazon
 	CloudVendorAWS CloudVendorType = "Amazon"
 	// CloudVendorAzure Azure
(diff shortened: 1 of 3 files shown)


pkg/apis/internal.open-cluster-management.io/v1beta1/clusterinfo_types.go
```diff
@@ -161,6 +161,10 @@ type CloudVendorType string
 const (
 	// CloudVendorIBM IBM
 	CloudVendorIBM CloudVendorType = "IBM"
+	// CloudVendorIBMZ IBM s360x
+	CloudVendorIBMZ CloudVendorType = "IBM Z"
```
qiujian16, line 165: IBMZPlatform?

pkg/apis/internal.open-cluster-management.io/v1beta1/clusterinfo_types.go
```diff
@@ -161,6 +161,10 @@ type CloudVendorType string
...
+	// CloudVendorIBMP IBM Power
+	CloudVendorIBMP CloudVendorType = "IBM Power"
```
qiujian16, line 167: IBMPowerPlatform

qiujian16 (commented):
i think we also need to handle clusterclaim

pkg/apis/internal.open-cluster-management.io/v1beta1/clusterinfo_types.go
zhiweiyin318: done

pkg/apis/internal.open-cluster-management.io/v1beta1/clusterinfo_types.go
zhiweiyin318: done

zhiweiyin318:
> i think we also need to handle clusterclaim

done

PR #313, opened 2021-03-18
fix issue that cannot get clusterview using verb *

zhiweiyin318:
fix issue that cannot get clusterview resources using verb *
fix typo


qiujian16 (commented):
/approve
/lgtm

PR #314, opened 2021-03-18
upgrade crd to v1

zhiweiyin318:
1. upgrade crd to v1
2. delete denynamespace webhook
3. fix conversion-gen path issue
story: https://github.com/open-cluster-management/backlog/issues/10443


deploy/foundation/hub/resources/crds/cluster.open-cluster-management.io_mirroredmanagedclusters.yaml
```diff
@@ -1,208 +1,172 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
qiujian16, line 3: I think we can delete this crd also, no one is using it.

cmd/webhook/app/server.go
```diff
@@ -95,7 +88,7 @@ func (a *admissionHandler) serve(w io.Writer, r *http.Request, admit admitFunc)
 func (a *admissionHandler) mutateResource(ar v1beta1.AdmissionReview) *v1beta1.AdmissionResponse {
```
qiujian16, line 88: change to v1.AdmissioinReview

cmd/webhook/app/server.go
```diff
@@ -95,7 +88,7 @@ func (a *admissionHandler) serve(w io.Writer, r *http.Request, admit admitFunc)
...
 	klog.V(2).Info("mutating custom resource")
 	raw := ar.Request.Object.Raw
-	crd := apiextensionsv1beta1.CustomResourceDefinition{}
+	crd := apiextensionsv1.CustomResourceDefinition{}
```
qiujian16, line 91: I am not sure why we use CustomResourceDefinition here. It seems to be an old bug.

pkg/apis/inventory/v1alpha1/baremetalasset_types.go
```diff
@@ -59,6 +59,7 @@ type BareMetalAssetSpec struct {
 	Role Role `json:"role,omitempty"`
 
 	// ClusterDeployment which the asset belongs to.
+	// +kubebuilder:pruning:PreserveUnknownFields
```
qiujian16, line 62: this seems to be weird, i need to check whether we do not use ObjectReference.

deploy/foundation/hub/resources/crds/cluster.open-cluster-management.io_mirroredmanagedclusters.yaml
zhiweiyin318: deleted

cmd/webhook/app/server.go
zhiweiyin318: done

cmd/webhook/app/server.go
qiujian16: let's update this part as a followup

pkg/apis/inventory/v1alpha1/baremetalasset_types.go
qiujian16: Let's update this as a followup

qiujian16 (commented):
/approve
/lgtm

PR #320, opened 2021-04-08
update clusterrole 

ldpliu:


pkg/utils/role.go
```diff
@@ -102,8 +102,12 @@ func ApplyClusterRoleBinding(ctx context.Context, client client.Client, required
 }
 
 //managedcluster admin role
-func GenerateClusterRoleName(clusterName string) string {
-	return "open-cluster-management:admin:" + clusterName
+func GenerateClusterRoleName(clusterName, role string) string {
+	return "open-cluster-management:" + role + ":" + clusterName
```
qiujian16, line 106: it is better to us fmt.Sprintf

pkg/controllers/clusterset/clusterrole/rbac.go
```diff
@@ -36,6 +37,10 @@ func buildAdminRoleRules(clustersetName string) []rbacv1.PolicyRule {
 			Groups(managedclusterGroup).
 			Resources("managedclusters").
 			RuleOrDie(),
+		clusterrbac.NewRule("update").
+			Groups(registerGroup).
+			Resources("managedclusters/accept").
```
qiujian16, line 42: this seems suspicious. I do not think admin for a clusterset can accept ANY cluster join

qiujian16: @ldpliu please add a todo here so we handle  this issue in another pr.

qiujian16 (commented):
/approve
/lgtm

PR #327, opened 2021-04-20
acstract cache

ldpliu:


ray-harris:
The OpenShift Prow squad is going to help debug why the sonar cloud job is failing. The check completes successfully, but Prow doesn't notice that it's done and times out after two hours. Please leave this PR as is so they can re-run the sonar cloud job. Thanks

PR #330, opened 2021-04-21
update OCPDistributionInfo

zhiweiyin318:
story: https://github.com/open-cluster-management/backlog/issues/11698

### DIFF
--- a/pkg/apis/internal.open-cluster-management.io/v1beta1/clusterinfo_types.go
+++ b/pkg/apis/internal.open-cluster-management.io/v1beta1/clusterinfo_types.go
@@ -29,14 +29,17 @@ type ClusterInfoStatus struct {
 	Version string `json:"version,omitempty"`
 
 	// KubeVendor describes the kubernetes provider of the managed cluster.
+	// Deprecated in release 2.3 and will be removed in the future. Use clusterClaim platform.open-cluster-management.io instead.
 	// +optional
 	KubeVendor KubeVendorType `json:"kubeVendor,omitempty"`
 
 	// CloudVendor describes the cloud provider for the managed cluster.
+	// Deprecated in release 2.3 and will be removed in the future. Use clusterClaim product.open-cluster-management.io instead.
 	// +optional
 	CloudVendor CloudVendorType `json:"cloudVendor,omitempty"`
 
-	// ClusterID is the identifier of managed cluster
+	// ClusterID is the identifier of managed cluster.
+	// Deprecated in release 2.3 and will be removed in the future. Use clusterClaim id.openshift.io instead.
 	// +optional
 	ClusterID string `json:"clusterID,omitempty"`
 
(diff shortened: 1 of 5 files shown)


pkg/apis/internal.open-cluster-management.io/v1beta1/clusterinfo_types.go
```diff
@@ -29,22 +29,26 @@ type ClusterInfoStatus struct {
 	Version string `json:"version,omitempty"`
 
 	// KubeVendor describes the kubernetes provider of the managed cluster.
+	// Deprecated in release 2.3. Use clusterClaim platform.open-cluster-management.io instead.
```
qiujian16, line 32: add  a comment that this is to be removed in the future

pkg/apis/internal.open-cluster-management.io/v1beta1/clusterinfo_types.go
```diff
@@ -112,16 +116,84 @@ type DistributionInfo struct {
...
+	// Deprecated in release 2.3. Use clusterClaim version.openshift.io instead.
 	Version string `json:"version,omitempty"`
-	// availableUpdates contains the list of update versions that are appropriate for the manage cluster.
+
+	// AvailableUpdates contains the list of update versions that are appropriate for the manage cluster.
+	// Deprecated in release 2.3. Use VersionAvailableUpdates instead.
 	AvailableUpdates []string `json:"availableUpdates,omitempty"`
-	// Desiredversion represents the desired version of the ocp cluster
+
+	// DesiredVersion is the version that the cluster is reconciling towards
```
qiujian16, line 176: also deprecate this?

zhiweiyin318: need add the whole fileds of [Desired](https://github.com/openshift/api/blob/a3c8fa53e01bc4442b687d3aca1571692f7676ce/config/v1/types_cluster_version.go#L87 )  to the clusterinfo if we deprecate this. I think `DesiredVersion` is enough.

pkg/apis/internal.open-cluster-management.io/v1beta1/clusterinfo_types.go
zhiweiyin318: done

pkg/klusterlet/clusterinfo/clusterinfo_controller.go
```diff
@@ -409,36 +409,79 @@ func (r *ClusterInfoReconciler) getOCPDistributionInfo() (clusterv1beta1.OCPDist
...
 	for _, update := range availableUpdates {
-		availableVersion, _, err := unstructured.NestedString(update.(map[string]interface{}), "version")
+		versionUpdate := clusterv1beta1.OCPVersionRelease{}
+		versionUpdate.Version, _, err = unstructured.NestedString(update.(map[string]interface{}), "version")
 		if err != nil {
-			klog.Errorf("failed to get OCP cluster version in latest history of status: %v", err)
-			continue
+			klog.Errorf("failed to get version in availableUpdates of clusterVersion: %v", err)
+		}
+		versionUpdate.Image, _, err = unstructured.NestedString(update.(map[string]interface{}), "image")
```
qiujian16, line 424: do we import openshift/api?  if yes, we should render to typed api instead  of using unstructured. Using nested could be erroneous.

zhiweiyin318: no, I tried to import openshift/api, but conflicted with the old version imported in hive.

qiujian16 (commented):
/approve
/lgtm

PR #334, opened 2021-04-26
Query prom to get socket/core

qiujian16:
https://github.com/open-cluster-management/backlog/issues/11105

still need unit test.

/hold


qiujian16:
unit test added. I do not think we could run e2e for this. but I do verify this on a bm cluster

/unhold
/assign @zhiweiyin318 
/assign @itdove 

we also need to add an additional rbac permission for the agent to update managedcluster status

PR #335, opened 2021-04-28
fix bug when replica is bigger than one

ldpliu:


cmd/controller/app/server.go
```diff
@@ -182,19 +182,31 @@ func Run(o *options.ControllerRunOptions, stopCh <-chan struct{}) error {
 		klog.Errorf("unable to setup gc reconciler: %v", err)
 		return err
 	}
+	go func() {
+		for {
+			select {
```
qiujian16, line 187: why  you need a for and select here? 

can we just block on <-mgr.Elected() and then run anything else?

qiujian16 (commented):
/lgtm

could you also change the agent in the same way?

ldpliu:
> /lgtm
> 
> could you also change the agent in the same way?

Ok, i will log a new task to handle this

PR #336, opened 2021-04-28
Add cpu_worker into capacity

qiujian16:


qiujian16:
https://github.com/open-cluster-management/backlog/issues/12025

qiujian16:
also fix node update lagging here.

PR #342, opened 2021-05-07
Remove cpu_worker/socket etc

qiujian16:
/hold

need to confirm with @itdove

### DIFF
--- a/pkg/controllers/clusterinfo/capacity_controller.go
+++ b/pkg/controllers/clusterinfo/capacity_controller.go
@@ -17,11 +17,9 @@ import (
 )
 
 const (
-	resourceCore         clusterv1.ResourceName = "core"
 	resourceSocket       clusterv1.ResourceName = "socket"
 	resourceCoreWorker   clusterv1.ResourceName = "core_worker"
 	resourceSocketWorker clusterv1.ResourceName = "socket_worker"
-	resourceCPUWorker    clusterv1.ResourceName = "cpu_worker"
 )
 
 type CapacityReconciler struct {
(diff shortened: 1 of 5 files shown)


pkg/controllers/clusterinfo/capacity_controller.go
```diff
@@ -62,25 +60,16 @@ func (r *CapacityReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
-	socketTotalCapacity := *resource.NewQuantity(int64(0), resource.DecimalSI)
 	socketWorkerCapacity := *resource.NewQuantity(int64(0), resource.DecimalSI)
-	coreTotalCapacity := *resource.NewQuantity(int64(0), resource.DecimalSI)
 	coreWorkerCapacity := *resource.NewQuantity(int64(0), resource.DecimalSI)
 	for _, node := range nodes {
-		socketTotalCapacity.Add(node.Capacity[resourceSocket])
-		coreTotalCapacity.Add(node.Capacity[resourceCore])
 		if isWorker(node) {
-			cpuWorkerCapacity.Add(node.Capacity[clusterv1.ResourceCPU])
+			coreWorkerCapacity.Add(node.Capacity[clusterv1.ResourceCPU])
```
itdove, line 67: I know that CPU/Core are the same but why you usse the lusterv1.ResourceCPU] to fill the core? isn't it confusing? Don't we have a capacity for the Core?
Ok got it... this is to optimize the api calls and so improve performance.

itdove (approved):
Just made a comment but now I understand the reason why you did like that (using cpu for core)

PR #353, opened 2021-05-17
update webhook qps and burst

ldpliu:
https://github.com/open-cluster-management/backlog/issues/12053


qiujian16:
do you  have a test result  on whether it will  enhance performance?

ldpliu:
> do you have a test result on whether it will enhance performance?

yes,
Before the change, when I create 100 managedclusters concurrently, many request will fail. 
And after the change, I can create 200 mangedclusters concurrently.

PR #356, opened 2021-05-20
add ocp version to managedcluster label

ldpliu:
https://github.com/open-cluster-management/backlog/issues/12633

### DIFF
--- a/pkg/controllers/clusterinfo/autodetect_controller.go
+++ b/pkg/controllers/clusterinfo/autodetect_controller.go
@@ -5,6 +5,8 @@ import (
 	"reflect"
 
 	clusterinfov1beta1 "github.com/open-cluster-management/multicloud-operators-foundation/pkg/apis/internal.open-cluster-management.io/v1beta1"
+	clusterclaims "github.com/open-cluster-management/multicloud-operators-foundation/pkg/klusterlet/clusterclaim"
+
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


pkg/controllers/clusterinfo/autodetect_controller.go
```diff
@@ -85,6 +86,13 @@ func (r *AutoDetectReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		needUpdate = true
 	}
 
+	if clusterInfo.Status.DistributionInfo.Type == clusterinfov1beta1.DistributionTypeOCP && clusterInfo.Status.DistributionInfo.OCP.Version != "" {
```
qiujian16, line 89: it needs ut. And i think we already collect it in cluster claim? why not update the label based on claim?

pkg/controllers/clusterinfo/autodetect_controller.go
```diff
@@ -85,6 +88,13 @@ func (r *AutoDetectReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		needUpdate = true
 	}
 
+	for _, claim := range cluster.Status.ClusterClaims {
+		if claim.Name == clusterclaims.ClaimOpenshiftVersion {
+			labels[OCPVersion] = claim.Value
+			needUpdate = true
```
qiujian16, line 94: will it update every time?

qiujian16 (commented):
/lgtm

PR #358, opened 2021-05-26
enable the junit report

hchenxa:
Signed-off-by: hchenxa <huichen@redhat.com>

The PR used to generate the junit result when we set the environment variable `JUNIT_REPORT_FILE` exported.

### DIFF
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -191,6 +191,28 @@ func GetHostFromClientConfig() (string, error) {
 	return clientCfg.Host, nil
 }
 
+func GetManagedCluster(dynamicClient dynamic.Interface, clusterName string) (*unstructured.Unstructured, error) {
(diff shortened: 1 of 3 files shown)


test/e2e/util/util.go
```diff
@@ -191,6 +191,28 @@ func GetHostFromClientConfig() (string, error) {
 	return clientCfg.Host, nil
 }
 
+func GetManagedCluster(dynamicClient dynamic.Interface, clusterName string) (*unstructured.Unstructured, error) {
+	cluster, err := dynamicClient.Resource(ManagedClusterGVR).Get(context.TODO(), clusterName, metav1.GetOptions{})
+	if err != nil {
```
qiujian16, line 196: this is to allow  running e2e separately against ACM. 

I would want this to use clusterClient  instead of dynamicClient  and also other  cluster related func here as  a follow up @zhiweiyin318 @ldpliu 

/apprrove

PR #361, opened 2021-06-01
add qps and burst for agent

ldpliu:
https://github.com/open-cluster-management/backlog/issues/12579

### DIFF
--- a/pkg/utils/rest/kubecontrol.go
+++ b/pkg/utils/rest/kubecontrol.go
@@ -65,6 +65,10 @@ func NewKubeControl(mapper meta.RESTMapper, config *rest.Config) *KubeControl {
 	}
 }
 
+// The resourceOrKindArg should be in the format of {resource}.{version}.{group}, or {resource}.
+// If it is {resource}.{group}, it will be mistakenly parsed to {resource: {resource}, version: {group or part of group if there is "." in group} }
+// which will trigger the reload of restmapper and introduce performance degradation.
+// TODO: refactor restmapper to avoid frequent reload.
 func MappingFor(mapper meta.RESTMapper, resourceOrKindArg string) (*meta.RESTMapping, error) {
 	fullySpecifiedGVR, groupResource := schema.ParseResourceArg(resourceOrKindArg)
 	gvk := schema.GroupVersionKind{}
--- a/cmd/agent/agent.go
+++ b/cmd/agent/agent.go
@@ -4,9 +4,10 @@ package main
 
 import (
 	"context"
-	openshiftclientset "github.com/openshift/client-go/config/clientset/versioned"
 	"os"
 	"time"
+
+	openshiftclientset "github.com/openshift/client-go/config/clientset/versioned"
 
 	"github.com/open-cluster-management/addon-framework/pkg/lease"
 	clusterclientset "github.com/open-cluster-management/api/client/cluster/clientset/versioned"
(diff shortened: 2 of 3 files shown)


qiujian16:
would you elaborate why it is qps that affect managedclusterview? Is there a lot of managedclusterview for a certain cluster?

ldpliu:
> would you elaborate why it is qps that affect managedclusterview? Is there a lot of managedclusterview for a certain cluster?

I reproduceed the problem and change managedcluster client QPS can fix this proble. 
if not, this line will take a long time. https://github.com/open-cluster-management/multicloud-operators-foundation/blob/aa4711bbfbc7d65c9bdb3f659aad5faba5d8eda1/pkg/utils/rest/kubecontrol.go#L72

qiujian16:
This is weird, i think kindfor will get gvk from local cache...did you change the lazy mode to false?

ldpliu:
> This is weird, i think kindfor will get gvk from local cache...did you change the lazy mode to false?

I changed it and it did not take affect

qiujian16:
I would want to understand more on why this will take long time. And will it impact work agent @elgnay? since we use the similar rest mapper

elgnay:
> I would want to understand more on why this will take long time. And will it impact work agent @elgnay? since we use the similar rest mapper

I'll look into it and see if we can find something interesting.

ldpliu:
I use this view, 
`apiVersion: view.open-cluster-management.io/v1beta1
kind: ManagedClusterView
metadata:
  name: 38e5b5d5cbdeb20187b3615840594a56d1d074ac-3
  namespace: local-cluster
spec:
  scope:
    name: local-cluster
    resource: managedclusters.cluster.open-cluster-management.io`
it looks the resource can not be found in local-cache(`err:no matches for open-cluster-management.io/cluster, Resource=managedclusters`), so it need to resync the cache. But i am not sure why it can not be found in local-cache

elgnay:
The mapper cache can be hit when `.spec.scope.resource` of `ManagedClusterView` is set to `managedclusters`.
while the cache will be reloaded if user set `.spec.scope.resource` with values, like `managedclusters.cluster.open-cluster-management.io`.

https://github.com/open-cluster-management/multicloud-operators-foundation/blob/0c5034a24e4b2259f758aaee464117818362a2a3/pkg/utils/rest/kubecontrol.go#L68-L76

In the above function, `mapper.KindFor()` will be called at most twice with different parameters.  When invoking `MappingFor()` with `managedclusters.cluster.open-cluster-management.io` as resourceOrKindArg, 
1. Go to the first call of `mapper.KindFor()`
2. No kind is found in the mapper cache; 
3. And the mapper reloads the cache;
4. Still no kind is found in the updated mapper cached either;
5. Fall back to the second call of `mapper.KindFor()`;
6. Kind is found;

It seems the `DynamicRESTMapper` is not a good choice for us. We need to control when the mapper cache should be reloaded by our own code.

ldpliu:
Same as @elgnay said, in .spec.scope.resource of ManagedClusterView, if its `managedclusters.cluster.open-cluster-management.io`, it will trigger the restmap sync logic. It will send many request and will be slow respense. So we need to improve qps to handle this case. 
@qiujian16 should we cherry-pick this pr to 2.2 ?

pkg/utils/rest/kubecontrol.go
```diff
@@ -65,6 +65,10 @@ func NewKubeControl(mapper meta.RESTMapper, config *rest.Config) *KubeControl {
 	}
 }
 
+// TODO:
+// If the resourceOrKindArg format is resource.group, the schema.ParseResourceArg return value
```
qiujian16, line 69: this is not quite clear.

it might be

The resourceOrKindArg should be in the format of {resource}.{version}.{group}, or {resource}. If it is {resource}.{group}, it will be mistakenly parsed to {resource: {resource}, version: {group} } which will trigger the reload of restmapper and introduce performance degradation. 
TODO: refactor restmapper to avoid frequent reload.

PR #364, opened 2021-06-04
fix claimed cluster can not delete from clusterset

ldpliu:
https://github.com/open-cluster-management/backlog/issues/13011

### DIFF
--- a/pkg/controllers/clusterset/clusterclaim/clusterclaim_controller.go
+++ b/pkg/controllers/clusterset/clusterclaim/clusterclaim_controller.go
@@ -54,11 +54,47 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
-	if err = c.Watch(&source.Kind{Type: &hivev1.ClusterClaim{}},
-		&handler.EnqueueRequestForObject{}); err != nil {
+	err = c.Watch(&source.Kind{Type: &hivev1.ClusterClaim{}},
+		&handler.EnqueueRequestForObject{})
+	if err != nil {
 		return err
 	}
 
+	//watch clusterdeployment related claim
+	err = c.Watch(
+		&source.Kind{Type: &hivev1.ClusterDeployment{}},
+		handler.EnqueueRequestsFromMapFunc(
+			handler.MapFunc(func(a client.Object) []reconcile.Request {
+				clusterDeployment, ok := a.(*hivev1.ClusterDeployment)
+				if !ok {
+					// not a ClusterDeployment, returning empty
+					klog.Error("ClusterDeployment handler received non-ClusterDeployment object")
+					return []reconcile.Request{}
+				}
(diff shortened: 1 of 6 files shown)


pkg/controllers/clusterset/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -62,28 +63,31 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 }
 
 func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	returnResult := reconcile.Result{
+		RequeueAfter: time.Second * 5,
```
qiujian16, line 67: why do we need to requeue?

test/e2e/clusterset_test.go
```diff
@@ -275,6 +275,55 @@ var _ = ginkgo.Describe("Testing ManagedClusterSet", func() {
 				}
 				return false
 			}, eventuallyTimeout, eventuallyInterval).Should(gomega.BeTrue())
+
+			//cliam the clusterdeployment and remove from set
+			gomega.Eventually(func() error {
```
qiujian16, line 280: this is a good test case

pkg/controllers/clusterset/clusterdeployment/clusterdeployment_controller.go
ldpliu: because we always need to sync the clusterdeloyment lablel with clusterpool if it's not claimed, if the clusterpool's label changed, we need to sync the change from requeue.

qiujian16: why not watch clusterpool and reconcile all related clusterdeployment?

ldpliu: done

pkg/controllers/clusterset/clusterclaim/clusterclaim_controller.go
```diff
@@ -54,11 +54,47 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+				}
+				var requests []reconcile.Request
+				for _, clusterclaim := range clusterclaims.Items {
+					//If clusterclaims is not related to this clusterDeployment, ignore it
+					if clusterclaim.Spec.Namespace != clusterDeployment.Namespace {
+						continue
+					}
+					requests = append(requests, reconcile.Request{
+						NamespacedName: types.NamespacedName{
+							Name:      clusterclaim.Spec.Namespace,
```
qiujian16, line 87: name or namespace?

ldpliu: it should be namespace, it means clusterdeployment namespace(same as name)

PR #366, opened 2021-06-06
reduce clusterinfo update

zhiweiyin318:
related issue: https://github.com/open-cluster-management/backlog/issues/12991

### DIFF
--- a/pkg/klusterlet/clusterinfo/clusterinfo_controller.go
+++ b/pkg/klusterlet/clusterinfo/clusterinfo_controller.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	openshiftclientset "github.com/openshift/client-go/config/clientset/versioned"
 	"net"
+	"sort"
+	"time"
 
 	clusterv1alpha1informer "github.com/open-cluster-management/api/client/cluster/informers/externalversions/cluster/v1alpha1"
 	clusterv1alpha1lister "github.com/open-cluster-management/api/client/cluster/listers/cluster/v1alpha1"
@@ -129,6 +131,15 @@ func (r *ClusterInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			Reason:  clusterv1beta1.ReasonManagedClusterInfoSyncedFailed,
 			Message: errors.NewAggregate(errs).Error(),
 		}
+	}
+
+	// check if distributionInfo is Updated.
+	// sort the slices in distributionInfo to make it comparable using DeepEqual if not update.
+	infoUpdated := distributionInfoUpdated(&clusterInfo.Status.DistributionInfo, &newStatus.DistributionInfo)
+
+	// need to sync ocp ClusterVersion info every 5 min since do not watch it.
+	if !infoUpdated && equality.Semantic.DeepEqual(newStatus, clusterInfo.Status) {
+		return ctrl.Result{RequeueAfter: time.Minute * 5}, nil
 	}
 
 	meta.SetStatusCondition(&newStatus.Conditions, newSyncedCondition)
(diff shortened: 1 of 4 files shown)


pkg/klusterlet/clusterinfo/clusterinfo_controller.go
```diff
@@ -388,6 +398,72 @@ func (r *ClusterInfoReconciler) getDistributionInfo() (clusterv1beta1.Distributi
...
+	}
+	return false
+}
+
+func ocpDistributionInfoUpdated(old, new *clusterv1beta1.OCPDistributionInfo) bool {
+	if len(old.AvailableUpdates) != len(new.AvailableUpdates) {
+		return true
+	}
+
+	var sortedAvailableUpdates []string
```
qiujian16, line 414: I think you could use  similar of this https://github.com/open-cluster-management/work/blob/main/pkg/spoke/controllers/appliedmanifestcontroller/appliedmanifestwork_controller.go#L167-L180 to sort?

zhiweiyin318: done

pkg/klusterlet/clusterinfo/clusterinfo_controller.go
```diff
@@ -388,6 +399,52 @@ func (r *ClusterInfoReconciler) getDistributionInfo() (clusterv1beta1.Distributi
...
+	}
+	return false
+}
+
+func ocpDistributionInfoUpdated(old, new *clusterv1beta1.OCPDistributionInfo) bool {
+	if len(old.AvailableUpdates) != len(new.AvailableUpdates) {
+		return true
+	}
+	if old.AvailableUpdates != nil && new.AvailableUpdates != nil {
+		sort.SliceStable(old.AvailableUpdates, func(i, j int) bool { return old.AvailableUpdates[i] < old.AvailableUpdates[j] })
```
qiujian16, line 415: Why do we need to sort the old one? I think it  is sorted already in the last update

zhiweiyin318: my understanding is that we can't guarantee that the old one is sorted  after marshal/Unmarshal. 
use nodeStatus as an example, the original issue happened when nodeStatus in clusterinfo are different from the ones got from nodeList (slice order is different) although they have been updated using the last ones got from nodeList.

qiujian16: but the old one is fetched from managedclusterinfo which should be sorted before write,  right?

zhiweiyin318: verified that the clusterinfo is not out of order. deleted the part to sort old one.

pkg/klusterlet/clusterinfo/clusterinfo_controller.go
```diff
@@ -388,6 +399,47 @@ func (r *ClusterInfoReconciler) getDistributionInfo() (clusterv1beta1.Distributi
...
+	}
+	return false
+}
+
+func ocpDistributionInfoUpdated(old, new *clusterv1beta1.OCPDistributionInfo) bool {
+	if len(old.AvailableUpdates) != len(new.AvailableUpdates) {
+		return true
+	}
+	if len(new.AvailableUpdates) > 1 {
+		sort.SliceStable(new.AvailableUpdates, func(i, j int) bool { return new.AvailableUpdates[i] < new.AvailableUpdates[j] })
```
qiujian16, line 415: how about
```go
func ocpDistributionInfoUpdated(old, new *clusterv1beta1.OCPDistributionInfo) bool {
  sort.SliceStable(new.AvailableUpdates, func(i, j int) bool { return new.AvailableUpdates[i] < new.AvailableUpdates[j] })
   sort.SliceStable(new.VersionAvailableUpdates, func(i, j int) bool {			return new.VersionAvailableUpdates[i].Version &lt; new.VersionAvailableUpdates[j].Version		})
  sort.SliceStable(new.VersionHistory, func(i, j int) bool { return new.VersionHistory[i].Version < new.VersionHistory[j].Version })
  return !equality.Semantic.DeepEqual(old, new)
}
```

zhiweiyin318: yes, this looks like more clear. 
 I checked that sort.SliceStable can handle empty slice.
done.

pkg/klusterlet/clusterinfo/clusterinfo_controller_test.go
```diff
@@ -499,3 +498,174 @@ func newConfigV1Client(version string, failingStatus bool) openshiftclientset.In
...
+						{
+							Image:    "quay.io/openshift-release-dev/ocp-release@sha256:4d048ae1274d11c49f9b7e70713a072315431598b2ddbb512aee4027c422fe3e",
+							State:    "Completed",
+							Verified: false,
+							Version:  "4.5.13",
+						},
+					},
+				},
+			},
+		},
```
qiujian16, line 664: can you add another two case
1. old is nil
2.  new is nil

zhiweiyin318: done

PR #367, opened 2021-06-14
fix webhook cert rotation

zhiweiyin318:
fix issue: https://github.com/open-cluster-management/backlog/issues/13329
add e2e cases for webhook
refactor some e2e cases using client API.


qiujian16 (commented):
/lgtm

PR #372, opened 2021-06-18
change rest mapper

ldpliu:
https://github.com/open-cluster-management/backlog/issues/12776


ldpliu:
@anthonyfister With this fix, we can reduce many api calls and could have a more consolidated fix for https://github.com/open-cluster-management/backlog/issues/12776. so it's better to merge this pr to 2.2.5.

qiujian16:
/hold
/approve
/lgtm

unhold this when we agree on that this should be backported in 2.2.5

PR #374, opened 2021-06-21
Skip updating user and group ID for appsubs

philipwu08:
https://github.com/open-cluster-management/backlog/issues/9533

Signed-off-by: Philip Wu <phwu@redhat.com>

### DIFF
--- a/cmd/webhook/app/options/options.go
+++ b/cmd/webhook/app/options/options.go
@@ -12,21 +12,23 @@ import (
 
 // Config contains the server (the webhook) cert and key.
 type Options struct {
-	CertFile       string
-	KeyFile        string
-	KubeConfigFile string
-	QPS            float32
-	Burst          int
+	CertFile              string
+	KeyFile               string
+	KubeConfigFile        string
+	QPS                   float32
+	Burst                 int
+	SkipOverwriteUserList []string
 }
 
 // NewOptions constructs a new set of default options for webhook.
 func NewOptions() *Options {
 	return &Options{
-		KubeConfigFile: "",
-		CertFile:       "",
-		KeyFile:        "",
-		QPS:            100.0,
-		Burst:          200,
+		KubeConfigFile:        "",
+		CertFile:              "",
+		KeyFile:               "",
+		QPS:                   100.0,
+		Burst:                 200,
+		SkipOverwriteUserList: []string{"system:serviceaccount:open-cluster-management-agent-addon:klusterlet-addon-appmgr"},
 	}
 }
 
@@ -42,6 +44,8 @@ func (c *Options) AddFlags(fs *pflag.FlagSet) {
 		"Maximum QPS to the hub server from this webhook.")
 	fs.IntVar(&c.Burst, "max-burst", c.Burst,
 		"Maximum burst for throttle.")
+	fs.StringSliceVar(&c.SkipOverwriteUserList, "skip-overwrite-user-list", c.SkipOverwriteUserList,
+		"List of users to skip overwriting of user identity annotations.")
 }
 
 type certificateCacheEntry struct {
--- a/pkg/webhook/useridentity/mutatingWebhook.go
+++ b/pkg/webhook/useridentity/mutatingWebhook.go
@@ -5,19 +5,22 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"net/http"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
 	"github.com/mattbaird/jsonpatch"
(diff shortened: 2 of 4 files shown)


jnpacker:
/lgtm (i'm not in owners)

* @qiujian16 are you okay with the const, or have an alternative suggestion?
* testcase if there are already FVT available

pkg/webhook/useridentity/mutatingWebhook.go
```diff
@@ -92,6 +95,17 @@ func (a *AdmissionHandler) mutateResource(ar v1.AdmissionReview) *v1.AdmissionRe
 	}
 
 	annotations := obj.GetAnnotations()
+
+	// Do not change the userInfo if the object is being created by the AppMgr.
+	// This value is stamped onto the Subscription object by Appmgr (Only for subscriptions)
+	klog.V(2).Infof("Object kind:%s, User request:%s, expected:%s", obj.GetKind(), ar.Request.UserInfo.Username, UserIDforAppMgr)
+	if obj.GetKind() == "Subscription" && ar.Request.UserInfo.Username == UserIDforAppMgr {
```
qiujian16, line 102: I would rather have a flag to specify which identity should be skipped, and has UserIDforAppMgr as the default value.

philipwu08: Quijing, thanks for the review. This is all new to me. Can you give me more details about this flag? Are you asking this flag to be passed in the AdmissionRequest?

qiujian16: https://github.com/open-cluster-management/multicloud-operators-foundation/blob/main/cmd/webhook/app/options/options.go#L33 add a flag entry here

philipwu08: Quijian, thank you for the hint. Hopefully I got it right. Please review. Thanks.

cmd/webhook/app/options/options.go
```diff
@@ -12,21 +12,23 @@ import (
...
-		CertFile:       "",
-		KeyFile:        "",
-		QPS:            100.0,
-		Burst:          200,
+		KubeConfigFile:    "",
+		CertFile:          "",
+		KeyFile:           "",
+		QPS:               100.0,
+		Burst:             200,
+		UserSkipOverwrite: "system:serviceaccount:open-cluster-management-agent-addon:klusterlet-addon-appmgr",
```
jnpacker, line 31: Since we have this open, did you want to make it support a list of users?

philipwu08: It crossed my mind as well, not sure if we have a use-case for it. I'll work on having the changes ready, but not push the commit until Qiujian agrees.

qiujian16: lets make it a list

pkg/webhook/useridentity/mutatingWebhook.go
```diff
@@ -87,6 +86,7 @@ func (a *AdmissionHandler) serve(w io.Writer, r *http.Request, admit admitFunc)
 
 func (a *AdmissionHandler) mutateResource(ar v1.AdmissionReview) *v1.AdmissionResponse {
 	klog.V(2).Info("mutating custom resource")
```
qiujian16, line 88: We always use v(4) here

pkg/webhook/useridentity/mutatingWebhook.go
```diff
@@ -92,6 +95,14 @@ func (a *AdmissionHandler) mutateResource(ar v1.AdmissionReview) *v1.AdmissionRe
 	}
 
 	annotations := obj.GetAnnotations()
+
+	if ar.Request.UserInfo.Username == a.UserSkipOverwrite {
+		klog.V(2).Infof("Skip add user and group for resource: %+v, name: %+v", ar.Request.Resource.Resource, obj.GetName())
```
qiujian16, line 100: v(4) here also

pkg/webhook/useridentity/mutatingWebhook.go
philipwu08: Thanks Quijian. New commit has the requested changes. Not really sure where to put the "contains" functions - it probably belongs in the util package but couldn't find a good home for it.

qiujian16 (commented):
/approve
/lgtm
/hold
 
hold in case you want others to review it, otherwise feel free to cancel it. BTW, it is better to squash before merge it.

pkg/webhook/useridentity/mutatingWebhook.go
qiujian16: there is a ContainsString in util pkg.

qiujian16:
thanks, would you like to squash the commit?

philipwu08:
squashed -thanks Quijian!

PR #377, opened 2021-06-25
handle upgrade from 2.2.X to 2.3 to remove finalizer in clusterrole

ldpliu:
https://github.com/open-cluster-management/backlog/issues/13670

### DIFF
--- a/pkg/controllers/gc/garbagecollector_controller.go
+++ b/pkg/controllers/gc/garbagecollector_controller.go
@@ -5,8 +5,13 @@ import (
 	"time"
 
 	actionv1beta1 "github.com/open-cluster-management/multicloud-operators-foundation/pkg/apis/action/v1beta1"
+	"github.com/open-cluster-management/multicloud-operators-foundation/pkg/utils"
 	"k8s.io/apimachinery/pkg/api/meta"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	utilwait "k8s.io/apimachinery/pkg/util/wait"
+
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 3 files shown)


zhiweiyin318:
I don't think it is a good option to create a controller to cleanup the finalizer only for the upgrade case.
could you consider to finish the cleanup job once at the beginning or cleanup the finalizer when the clusterrole is reconciled.
cc @qiujian16

ldpliu:
> I don't think it is a good option to create a controller to cleanup the finalizer only for the upgrade case.
> could you consider to finish the cleanup job once at the beginning or cleanup the finalizer when the clusterrole is reconciled.
> cc @qiujian16

Currentlly, it's not a real controller, and it's a go-rotine and will only start on controller start and will finish after the finalizer cleaned.

pkg/controllers/gc/garbagecollector_controller.go
```diff
@@ -81,3 +87,47 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	return &CleanGarbageFinalizer{
+		kubeClient: kubeClient,
+		ctx:        ctx,
+		stopCh:     stopCh,
+	}
+}
+
+// start a routine to sync the clusterrolebinding periodically.
+func (c *CleanGarbageFinalizer) Run() {
+	go utilwait.Until(c.clean, 60*time.Second, c.stopCh)
```
qiujian16, line 109: if you run go outside, you do not need goroutine here. And you can use wait.PollImmediateUntil so no need to close channel in clean.

pkg/controllers/gc/garbagecollector_controller.go
```diff
@@ -81,3 +87,47 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	if err != nil {
+		return
+	}
+	allFinalizerRmoved := true
+	for _, clusterrole := range clusterRolebindingList.Items {
+		if !utils.ContainsString(clusterrole.GetFinalizers(), ClustersetFinalizerName) {
+			continue
+		}
+
+		klog.Infof("removing ClusterRoleBinding Finalizer in ClusterRole %v", clusterrole.Name)
```
qiujian16, line 123: V(4) please

qiujian16 (commented):
/approve
/lgtm

PR #381, opened 2021-06-30
empty cluster status when cluster is offline

zhiweiyin318:
related issue: https://github.com/open-cluster-management/backlog/issues/13596

### DIFF
--- a/pkg/controllers/clusterinfo/clusterinfo_controller.go
+++ b/pkg/controllers/clusterinfo/clusterinfo_controller.go
@@ -188,15 +188,21 @@ func (r *ClusterInfReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	}
 
 	// TODO: the conditions of managed cluster need to be deprecated.
-	newConditions := cluster.Status.Conditions
+	clusterConditions := cluster.Status.Conditions
+	newClusterInfo := clusterInfo.DeepCopy()
+
+	// empty node list in clusterInfo status when cluster is offline
+	if utils.ClusterIsOffLine(clusterConditions) {
+		newClusterInfo.Status.NodeList = []clusterinfov1beta1.NodeStatus{}
+	}
+	newClusterInfo.Status.Conditions = clusterConditions
 	syncedCondition := meta.FindStatusCondition(clusterInfo.Status.Conditions, clusterinfov1beta1.ManagedClusterInfoSynced)
 	if syncedCondition != nil {
-		newConditions = append(newConditions, *syncedCondition)
+		newClusterInfo.Status.Conditions = append(newClusterInfo.Status.Conditions, *syncedCondition)
 	}
 
-	if !reflect.DeepEqual(newConditions, clusterInfo.Status.Conditions) {
-		clusterInfo.Status.Conditions = newConditions
-		err = r.client.Status().Update(ctx, clusterInfo)
+	if !reflect.DeepEqual(newClusterInfo.Status, clusterInfo.Status) {
+		err = r.client.Status().Update(ctx, newClusterInfo)
 		if err != nil {
 			klog.Warningf("will reconcile since failed to update ManagedClusterInfo status %v, %v", cluster.Name, err)
 			return ctrl.Result{}, err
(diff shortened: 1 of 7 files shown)


pkg/utils/utils.go
```diff
@@ -29,3 +31,12 @@ func BuildKubeClient(kubeConfigPath string) (*kubernetes.Clientset, error) {
 func ResourceNamespacedName(resourceType, namespace, name string) string {
 	return fmt.Sprintf("%s/%s/%s", resourceType, namespace, name)
 }
+
+func ClusterIsOffLine(conditions []metav1.Condition) bool {
+	for _, sc := range conditions {
```
qiujian16, line 36: use meta.IsConditionTrue

pkg/controllers/clusterinfo/clusterinfo_controller.go
```diff
@@ -188,15 +188,21 @@ func (r *ClusterInfReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	}
 
 	// TODO: the conditions of managed cluster need to be deprecated.
-	newConditions := cluster.Status.Conditions
+	clusterConditions := cluster.Status.Conditions
+	newClusterInfo := clusterInfo.DeepCopy()
+
+	// empty clusterInfo status when cluster is offline
+	if utils.ClusterIsOffLine(clusterConditions) {
+		newClusterInfo.Status = clusterinfov1beta1.ClusterInfoStatus{}
```
qiujian16, line 196: I think we only empty the node list here.

pkg/utils/utils.go
zhiweiyin318: done

pkg/controllers/clusterinfo/clusterinfo_controller.go
zhiweiyin318: done

pkg/klusterlet/nodecollector/nodecollecter.go
```diff
@@ -108,6 +109,10 @@ func (r *resourceCollector) reconcile(ctx context.Context) {
 		klog.Errorf("Failed to get cluster: %v", err)
 	}
 
+	if utils.ClusterIsOffLine(clusterinfo.Status.Conditions) {
```
qiujian16, line 112: I do not think this is needed. Are you  saying cluster is offline but addon is  still running?

pkg/utils/utils.go
```diff
@@ -29,3 +32,7 @@ func BuildKubeClient(kubeConfigPath string) (*kubernetes.Clientset, error) {
 func ResourceNamespacedName(resourceType, namespace, name string) string {
 	return fmt.Sprintf("%s/%s/%s", resourceType, namespace, name)
 }
+
+func ClusterIsOffLine(conditions []metav1.Condition) bool {
+	return !meta.IsStatusConditionTrue(conditions, clusterapiv1.ManagedClusterConditionAvailable)
```
qiujian16, line 37: we should  only consider unknown  status

pkg/klusterlet/clusterinfo/clusterinfo_controller.go
```diff
@@ -72,6 +73,10 @@ func (r *ClusterInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	if utils.ClusterIsOffLine(clusterInfo.Status.Conditions) {
+		return ctrl.Result{RequeueAfter: time.Minute * 5}, nil
```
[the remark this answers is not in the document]
zhiweiyin318: we do not watch the upgrade info like channel/version of OCP, need to requeue to check these info.

pkg/klusterlet/nodecollector/nodecollecter.go
zhiweiyin318: yes, add this for this case.

pkg/utils/utils.go
zhiweiyin318: I found that we have false status for ManagedClusterConditionAvailable when managed cluster api server is Unavailable. https://github.com/open-cluster-management-io/registration/blob/e6435f32a3ed877b974f3299bd26db1101162f7b/pkg/spoke/managedcluster/healthcheck_controller.go#L96

zhiweiyin318: done

PR #388, opened 2021-07-30
add e2e for managedcluster ca

ldpliu:
https://github.com/open-cluster-management/backlog/issues/14136


test/e2e/managedcluster_test.go
```diff
@@ -25,5 +31,117 @@ var _ = ginkgo.Describe("Testing ManagedCluster", func() {
 				return nil
 			}, eventuallyTimeout, eventuallyInterval).ShouldNot(gomega.HaveOccurred())
 		})
+
+		ginkgo.It("Get CA from apiserver", func() {
+			//Create a fake secret for apiserver
+			fakesecretName := "fake-server-secret"
+			fakeSecret, err := e2eutil.CreateFakeTlsSecret(kubeClient, fakesecretName, utils.OpenshiftConfigNamespace)
```
qiujian16, line 38: why do we need a fake secret?

ldpliu: I need to fake secret as serving certs, and add it to apiserver to test get ca from apiserver.

zhiweiyin318: should cleanup the fake resource after the case.

ldpliu: sure

qiujian16:
/hold
/lgtm

@zhiweiyin318 leave to you to unhold.

test/e2e/managedcluster_test.go
```diff
@@ -25,5 +31,129 @@ var _ = ginkgo.Describe("Testing ManagedCluster", func() {
 				return nil
 			}, eventuallyTimeout, eventuallyInterval).ShouldNot(gomega.HaveOccurred())
 		})
+
+		ginkgo.It("Get CA from apiserver", func() {
```
zhiweiyin318, line 35: I think it will be better to start a new Context to include your ca cases, not in "Get ManagedCluster cpu worker capacity" Context.

test/e2e/managedcluster_test.go
```diff
@@ -25,5 +31,129 @@ var _ = ginkgo.Describe("Testing ManagedCluster", func() {
...
+					}
+				}
+				return false
+			}, eventuallyTimeout, eventuallyInterval).Should(gomega.BeTrue())
+
+			//rollback apiserver and delete secret
+			newApiserver.Spec.ServingCerts.NamedCertificates = []configv1.APIServerNamedServingCert{}
+			_, err = ocpClient.ConfigV1().APIServers().Update(context.TODO(), newApiserver, metav1.UpdateOptions{})
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+			err = kubeClient.CoreV1().Secrets(utils.OpenshiftConfigNamespace).Delete(context.TODO(), fakesecretName, metav1.DeleteOptions{})
```
zhiweiyin318, line 93: how about put create fake resource in Before , delete them in After ?

test/e2e/managedcluster_test.go
ldpliu: done

test/e2e/managedcluster_test.go
ldpliu: I think before and after will run in every case, the current resources are only used in this case

PR #391, opened 2021-08-12
upgrade hive to 1.1.12

ldpliu:
https://github.com/open-cluster-management/backlog/issues/14856


pkg/klusterlet/nodecollector/nodecollecter.go
```diff
@@ -254,12 +254,16 @@ func (r *resourceCollector) newPrometheusClient(caData []byte) (prometheusv1.API
...
+	httpTransport.TLSClientConfig = &tls.Config{RootCAs: r.certPool, MinVersion: tls.VersionTLS12}
 
+	roundTripper, err := transport.NewBearerAuthWithRefreshRoundTripper("", r.tokenFile, httpTransport)
+	if err != nil {
+		return nil, err
+	}
 	// read token from token files
 	client, err := prometheusapi.NewClient(prometheusapi.Config{
 		Address:      r.server,
-		RoundTripper: prometheusconfig.NewBearerAuthFileRoundTripper(r.tokenFile, transport),
```
zhiweiyin318, line 262: need verify if this change works.  update the ut and e2e to cover this.

ldpliu: we have e2e to cover this case: https://github.com/open-cluster-management/multicloud-operators-foundation/blob/d7bf1af18c9459f119767ae8026c35fbc80121f9/test/e2e/clusterinfos_test.go#L22

PR #401, opened 2021-09-14
fix voluerable lib dependencies

xuezhaojun:

Change to minimal version of autorest/adal so that it can pass the scan without changing other codes


xuezhaojun:
an e2e issue occurred but it doesn't seem caused by this PR

PR #411, opened 2021-10-18
fix the vendor is others for ROSA clusters

zhiweiyin318:
issue: https://github.com/open-cluster-management/backlog/issues/17180


qiujian16 (commented):
/approve
/lgtm

PR #412, opened 2021-10-21
Remove the sleep wait in config checker e2e test

xuezhaojun:


xuezhaojun:
Here, we also removed the entire config-checker e2e test case. 

The reason is config-checker is relying on the change of work-manager-hub-kubeconfig secret. After we change the content of this secret, the registration will reconcile this secret back quickly.

PR #429, opened 2021-12-15
add test for config checker（and delete old one）

xuezhaojun:


xuezhaojun:
The config-checker needs `livenessProbe` to restart the container, so integration-test still only can test half of the whole function.
So integration-test is on-process. But here we create a unit test for `ServeHealthProbes` to test config-checer's usage in work-manager.

PR #434, opened 2022-01-14
[backplane-1.0]GitHub org rename

zhujian7:


zhujian7:
cherry-pick https://github.com/stolostron/multicloud-operators-foundation/pull/429 to fix e2e problem.

PR #444, opened 2022-02-21
fix no openshift console in SNO

zhiweiyin318:
SNO has no openshift-console,  to return error will break the other clusterClaims collecting.

### DIFF
--- a/pkg/klusterlet/clusterclaim/clusterclaimer.go
+++ b/pkg/klusterlet/clusterclaim/clusterclaimer.go
@@ -343,41 +343,53 @@ func (c *ClusterClaimer) getClusterRegion() (string, error) {
 }
 
 // for OpenShift, read endpoint address from console-config in openshift-console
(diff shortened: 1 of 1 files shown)


pkg/klusterlet/clusterclaim/clusterclaimer.go
```diff
@@ -113,7 +113,6 @@ func (c *ClusterClaimer) List() ([]*clusterv1alpha1.ClusterClaim, error) {
 	version, clusterID, err := c.getOCPVersion()
 	if err != nil {
 		klog.Errorf("failed to get OCP version and clusterID, error: %v ", err)
-		return claims, err
```
skeeey, line 116: may be an aggregate error, then return this aggregate err at the end?

```
errs := appen(errs, err)
```

skeeey:
refer to https://github.com/stolostron/backlog/issues/20173

PR #447, opened 2022-03-03
use new set api

ldpliu:


pkg/cache/managedclusterset.go
```diff
@@ -54,6 +54,11 @@ func (c *ClusterSetCache) ListResources() (sets.String, error) {
 	}
 
 	for _, clusterSet := range clusterSets {
+		// ignore the non-legacy clusterset
+		selectorType := clusterSet.Spec.ClusterSelector.SelectorType
+		if len(selectorType) > 0 && selectorType != clusterv1beta1.LegacyClusterSetLabel {
```
zhiweiyin318, line 59: what if selectorType is empty?

PR #462, opened 2022-04-08
fix the addon is  failed to deploy on OCP311

zhiweiyin318:
issue:  https://github.com/stolostron/backlog/issues/21554

### DIFF
--- a/pkg/addon/addon.go
+++ b/pkg/addon/addon.go
@@ -45,7 +45,7 @@ type GlobalValues struct {
 }
 
 type Values struct {
-	IsOCP        bool         `json:"isOCP,omitempty"`
+	HasRoute     bool         `json:"hasRoute,omitempty"`
 	GlobalValues GlobalValues `json:"global,omitempty,omitempty"`
 }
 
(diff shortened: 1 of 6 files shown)


pkg/addon/addon.go
```diff
@@ -69,8 +69,9 @@ func NewGetValuesFunc(imageName string, imageRegistryClient imageregistry.Client
 		}
 
 		for _, claim := range cluster.Status.ClusterClaims {
-			if claim.Name == "product.open-cluster-management.io" && claim.Value == "OpenShift" {
-				addonValues.IsOCP = true
+			if claim.Name == "version.openshift.io" && claim.Value != "3" {
```
qiujian16, line 72: this is a danger claim, we cannot say it is OCP4 if it is not 3.

zhiweiyin318: the cluster is OCP if there is version.openshift.io clusterclaim.  only 2 options for the version, 3 or 4.x.x currently.

qiujian16: I would use HasRoute as flag instead using ISOCP4. Also add a comment in this section on how we detect whether route exists

zhiweiyin318: done

PR #463, opened 2022-04-12
Add backup label

qiujian16:
Signed-off-by: Jian Qiu <jqiu@redhat.com>

### DIFF
--- a/pkg/controllers/inventory/inventory_controller.go
+++ b/pkg/controllers/inventory/inventory_controller.go
@@ -15,6 +15,7 @@ import (
 	objectreferencesv1 "github.com/openshift/custom-resource-status/objectreferences/v1"
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	hiveinternalv1alpha1 "github.com/openshift/hive/apis/hiveinternal/v1alpha1"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 	inventoryv1alpha1 "github.com/stolostron/multicloud-operators-foundation/pkg/apis/inventory/v1alpha1"
 	bmaerrors "github.com/stolostron/multicloud-operators-foundation/pkg/controllers/inventory/errors"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


pkg/controllers/inventory/inventory_controller.go
```diff
@@ -342,15 +343,24 @@ func (r *ReconcileBareMetalAsset) checkAssetSecret(ctx context.Context, instance
...
 			return err
 		}
-		if err := r.client.Update(ctx, secret); err != nil {
-			klog.Errorf("Failed to update secret with OwnerReferences, %v", err)
-			return err
-		}
+	}
+
+	secretTypeLabel := map[string]string{
+		"cluster.open-cluster-management.io/credentials": "",
```
birsanv, line 356: These are the labels used by the Credentials showing up in the Credentials UI

I am not sure which one , I think the second label, will result in showing the secret in the Credentials UI, as a bma type. Do you have such a Credentials UI template ? Do you want those secrets to show up under the Credentials page ? If you do the 2 labels are good..

For backup purpose though, you can use either of the following secrets, both result in backing up the secrets having any of this labels: 
- cluster.open-cluster-management.io/type
- cluster.open-cluster-management.io/backup
If the label is added just for the purpose of backing up the secret ( no UI interaction ), then I suggest to use cluster.open-cluster-management.io/backup

birsanv (commented):
had a comment on the label used, please review

pkg/controllers/inventory/inventory_controller.go
```diff
@@ -342,15 +343,23 @@ func (r *ReconcileBareMetalAsset) checkAssetSecret(ctx context.Context, instance
...
 			return err
 		}
-		if err := r.client.Update(ctx, secret); err != nil {
-			klog.Errorf("Failed to update secret with OwnerReferences, %v", err)
-			return err
-		}
+	}
+
+	secretTypeLabel := map[string]string{
+		"cluster.open-cluster-management.io/backup": "",
```
birsanv, line 356: looks good; it's optional but if you want to tag the label to know it's used by a BMA asset, then you can add the value "bma" as you did initially with the other label

cluster.open-cluster-management.io/backup: "bma"

birsanv (commented):
one more comment, optional change; everything else looks good

PR #468, opened 2022-04-22
fix no route in ocp 311 and errors in log

zhiweiyin318:

1. fix issue: https://github.com/stolostron/backlog/issues/21853
OCP 311 has route, support to get log using route.

2. fix issue: https://github.com/stolostron/backlog/issues/21554
work-agent cannot update service if the type of service is changed in OCP 311, and return error 
`invalid: spec.ports[0].nodePort: Forbidden: may not be used when `type` is 'ClusterIP'`
this is an issue in old k8s version,  has been fixed in new one. https://github.com/kubernetes/kubectl/issues/221
the fix in addon is that deploy service until the product claim has value. will not update it in addon manifestwork.

3. return error for the interface of the imageRegisty Client to help debug.

### DIFF
--- a/pkg/addon/manifests/chart/templates/service.yaml
+++ b/pkg/addon/manifests/chart/templates/service.yaml
@@ -1,4 +1,5 @@
 ---
+{{- if not (eq .Values.product "") }}
 apiVersion: v1
 kind: Service
 metadata:
@@ -7,7 +8,7 @@ metadata:
   labels:
     component: "work-manager"
 spec:
-{{- if eq .Values.hasRoute false }}
+{{- if not (eq .Values.product "OpenShift") }}
   type: LoadBalancer
 {{- end }}
   ports:
@@ -17,3 +18,4 @@ spec:
       targetPort: 4443
   selector:
     component: "work-manager"
+{{- end }}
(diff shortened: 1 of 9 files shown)


pkg/addon/manifests/chart/templates/service.yaml
```diff
@@ -1,4 +1,5 @@
 ---
+{{- if not (eq .Values.product "") }}
```
qiujian16, line 2: not quite sure why this is needed, we do not create service when product is empty?

zhiweiyin318: the previous logic is that we create ClusterIP service firstly and then update the service when the claim is appended into the managedCluster by work-manager addon. 
for OCP 311, work-agent will meet error to update the service` invalid: spec.ports[0].nodePort: Forbidden: may not be used when type is 'ClusterIP'` because this issue https://github.com/kubernetes/kubectl/issues/221 has not fixed in old k8s.
current fix is that to create service until the product claim is appended to the managedCluster. will not need to update the service.

qiujian16: it means we will not create service in *ks clusters since we cannot know the product? is this desired?

zhiweiyin318: no, the product claim will be Others for *ks not empty.

PR #471, opened 2022-05-05
allow claimed cluster move to default set

ldpliu:
https://github.com/stolostron/backlog/issues/22227#event-6553624161

### DIFF
--- a/pkg/webhook/clusterset/validatingWebhook.go
+++ b/pkg/webhook/clusterset/validatingWebhook.go
@@ -48,6 +48,7 @@ var agentClusterOwnerRef = metav1.OwnerReference{
 
 const (
 	clusterSetLabel = "cluster.open-cluster-management.io/clusterset"
+	defaultSetName  = "default"
 )
 
 // validateResource validate:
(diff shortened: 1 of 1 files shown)


pkg/webhook/clusterset/validatingWebhook.go
```diff
@@ -163,6 +164,11 @@ func (a *AdmissionHandler) validateUpdateRequest(request *v1.AdmissionRequest) *
 			return rejectStatus
 		}
 
+		//For upgrade from 2.4 to 2.5, we need to move all clusters from empty set to "default" set.
+		if currentClusterSetName == defaultSetName {
+			return a.allowUpdate(request.UserInfo, originalClusterSetName, currentClusterSetName)
```
qiujian16, line 169: I think this should only be bypassed when the cluster is from the claim.  It should be like in line 188

```go
// the cluster is claimed from clusterpool, do not allow to update clustersetlabel.
if clusterDeployment == nil || clusterDeployment.Spec.ClusterPoolRef == nil {
	return a.allowUpdate(request.UserInfo, originalClusterSetName, currentClusterSetName)
}

if len(originalClusterSetName) == 0 && currentClusterSetName == defaultSetName {
    return a.allowUpdate(request.UserInfo, originalClusterSetName, currentClusterSetName)
}

rejectStatus.Result.Message = "Do not allow to update claimed cluster's clusterset."
return rejectStatus

```

pkg/webhook/clusterset/validatingWebhook.go
```diff
@@ -179,6 +180,11 @@ func (a *AdmissionHandler) validateUpdateRequest(request *v1.AdmissionRequest) *
 
 		// the cluster is claimed from clusterpool, do not allow to update clustersetlabel.
 		if clusterDeployment != nil && clusterDeployment.Spec.ClusterPoolRef != nil {
+			//For upgrade from 2.4 to 2.5, we need to move all clusters from empty set to "default" set.
+			if currentClusterSetName == defaultSetName {
```
zhiweiyin318, line 184: do we need to check if the old cluster has no clusterset here?

ldpliu: done

pkg/webhook/clusterset/validatingWebhook.go
ldpliu: done

PR #477, opened 2022-05-06
add test for cluster from clusterpool

ldpliu:


ldpliu:
> /hold will add e2e later

It's difficult to create a cluster without clusterset(registration will add default set to new cluster), so only add the ut test.
/unhold

PR #499, opened 2022-07-08
Add an option to control `sync labels to clusterclaims`

xuezhaojun:

Relate issue: https://github.com/stolostron/backlog/issues/23826

### DIFF
--- a/cmd/agent/agent.go
+++ b/cmd/agent/agent.go
@@ -244,13 +244,14 @@ func startManager(o *options.AgentOptions, ctx context.Context) {
 			ConfigV1Client: openshiftClient,
 			OauthV1Client:  osOauthClient,
 			Mapper:         restMapper,
+			EnableSyncLabelsToClusterClaims: o.EnableSyncLabelsToClusterClaims,
 		}
 
 		clusterClaimReconciler := clusterclaimctl.ClusterClaimReconciler{
-			Log:               ctrl.Log.WithName("controllers").WithName("ManagedClusterInfo"),
-			ClusterClient:     managedClusterClusterClient,
-			ClusterInformers:  clusterInformerFactory.Cluster().V1alpha1().ClusterClaims(),
-			ListClusterClaims: clusterClaimer.List,
+			Log:                             ctrl.Log.WithName("controllers").WithName("ManagedClusterInfo"),
+			ClusterClient:                   managedClusterClusterClient,
+			ClusterInformers:                clusterInformerFactory.Cluster().V1alpha1().ClusterClaims(),
+			ListClusterClaims:               clusterClaimer.List,
 		}
 
 		if err = actionReconciler.SetupWithManager(mgr); err != nil {
@@ -268,11 +269,9 @@ func startManager(o *options.AgentOptions, ctx context.Context) {
 			os.Exit(1)
 		}
 
-		if o.EnableSyncLabelsToClusterClaims == true {
-			if err = clusterClaimReconciler.SetupWithManager(mgr); err != nil {
-				setupLog.Error(err, "unable to create controller", "controller", "ClusterClaim")
-				os.Exit(1)
-			}
+		if err = clusterClaimReconciler.SetupWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create controller", "controller", "ClusterClaim")
+			os.Exit(1)
 		}
 
 		go kubeInformerFactory.Start(ctx.Done())
--- a/cmd/agent/agent.go
+++ b/cmd/agent/agent.go
@@ -238,20 +238,20 @@ func startManager(o *options.AgentOptions, ctx context.Context) {
 			AgentService:   o.AgentService,
 		}
 		clusterClaimer := clusterclaimctl.ClusterClaimer{
-			ClusterName:    o.ClusterName,
-			HubClient:      mgr.GetClient(),
-			KubeClient:     managedClusterKubeClient,
-			ConfigV1Client: openshiftClient,
-			OauthV1Client:  osOauthClient,
-			Mapper:         restMapper,
+			ClusterName:                     o.ClusterName,
+			HubClient:                       mgr.GetClient(),
+			KubeClient:                      managedClusterKubeClient,
+			ConfigV1Client:                  openshiftClient,
+			OauthV1Client:                   osOauthClient,
+			Mapper:                          restMapper,
 			EnableSyncLabelsToClusterClaims: o.EnableSyncLabelsToClusterClaims,
 		}
 
 		clusterClaimReconciler := clusterclaimctl.ClusterClaimReconciler{
-			Log:                             ctrl.Log.WithName("controllers").WithName("ManagedClusterInfo"),
-			ClusterClient:                   managedClusterClusterClient,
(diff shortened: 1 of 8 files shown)


cmd/controller/app/options/options.go
```diff
@@ -51,6 +53,8 @@ func (o *ControllerRunOptions) AddFlags(fs *pflag.FlagSet) {
 			"Enabling this will ensure there is only one active controller manager.")
 	fs.BoolVar(&o.EnableAddonDeploy, "enable-agent-deploy", o.EnableAddonDeploy,
 		"Enable deploy addon agent.")
+
+	fs.BoolVar(&o.EnableClusterClaims, "enable-clusterclaims", o.EnableClusterClaims, "Enable to create clusterclaims on the managed cluster")
```
qiujian16, line 57: do we want to call it enableClusterClaim or enableHostedMode? I think it is not just disable the clusterclaim, for reading core/socket, are those part still working in hosted mode?

xuezhaojun: Sure, let me check "reading core/socket" part of the code.

xuezhaojun:
Another issue for this PR from @elgnay is: if we set `EnableClusterClaim=false` for SD hub, are we expect SD not `read and write` or only not `write` cluterclaims?

Because in that case, if the customer hub imports the managed cluster first, and then SD hub can see clusterclaims in`managdcluster.status`.

But in the same case, if the SD hub imports the managed cluster first, at that time, it can not see any clusterclaims.
 
That is an inconsistent behavior and need to note down in our document. What do you think? @zhujian7

zhujian7:
I think we need more discussion about this: If the clusterclaim(ex, the `name` clusterclaim) is created for the custom hub, then will the SD hub read the "wrong" value?

cmd/controller/app/options/options.go
xuezhaojun: `name: cluster1`

to

`name: {"hub1": "cluster1", "hub2": "cluster2"}`

cmd/agent/agent.go
```diff
@@ -249,9 +249,11 @@ func startManager(o *options.AgentOptions, ctx context.Context) {
 			os.Exit(1)
 		}
 
-		if err = clusterClaimReconciler.SetupWithManager(mgr); err != nil {
-			setupLog.Error(err, "unable to create controller", "controller", "ClusterClaim")
-			os.Exit(1)
+		if o.EnableSyncLabelsToClusterClaims == true {
```
zhujian7, line 252: I am thinking if it is better to make the SyncLabelsToClusterClaims enable as default, and only disable it for the sd hub? if yes, this could be `o.DisableSyncLabelsToClusterClaims` ?

xuezhaojun: It is enabled by default now.

cmd/agent/app/options/options.go
```diff
@@ -8,48 +8,50 @@ import (
...
-		QPS:                  50,
-		Burst:                100,
+		LeaseDurationSeconds:            60,
+		MetricsAddr:                     ":8080",
+		KubeConfig:                      "",
+		HubKubeConfig:                   "/var/run/hub/kubeconfig",
+		ClusterName:                     "",
+		EnableLeaderElection:            true,
+		EnableImpersonation:             false,
+		EnableSyncLabelsToClusterClaims: true,
```
xuezhaojun, line 45: @zhujian7  Just like EnableLeaderElection, it's true by default.

pkg/addon/addon.go
```diff
@@ -56,6 +58,13 @@ func NewGetValuesFunc(imageName string) addonfactory.GetValuesFunc {
 		if err != nil {
 			return nil, err
 		}
+
+		// if cluster is hosed mode, the enableSyncLabelsToClusterClaims is false
```
zhujian7, line 62: should we check the addon deploy mode or the klusterlet deploy mode?

zhujian7: if the klusterlet is deployed in Hosted mode, but the work manager addon is deployed in Default mode(hypershift case in 2.5), will we enable or disable this functionality?

xuezhaojun: You're right, I only considered SD case before, here should based on addon annotation.

xuezhaojun: WIP

pkg/addon/addon.go
```diff
@@ -56,6 +58,13 @@ func NewGetValuesFunc(imageName string) addonfactory.GetValuesFunc {
 		if err != nil {
 			return nil, err
 		}
+
+		// if addon is hosed mode, the enableSyncLabelsToClusterClaims is false
+		enableSyncLabelsToClusterClaims := true
+		if value, ok := addon.GetAnnotations()[foundationutils.DeployAddonHostingKey]; ok && value != "" {
+			enableSyncLabelsToClusterClaims = false
```
xuezhaojun, line 65: @zhujian7  Please take a look here. We use addon annotation now.

pkg/addon/addon.go
xuezhaojun: Done.

pkg/utils/annotation.go
```diff
@@ -0,0 +1,10 @@
+package utils
+
+const (
+	// DeployModeAnnotationKey describe the klusterlet deploy mode when importing a managed cluster.
+	DeployModeAnnotationKey string = "import.open-cluster-management.io/klusterlet-deploy-mode"
```
zhujian7, line 5: do we still need this annotation?

pkg/utils/annotation.go
```diff
@@ -0,0 +1,10 @@
...
+	DeployModeValueHosted          = "Hosted"
+
+	// DeployAddonHostingKey
```
zhujian7, line 8: add some comments?

pkg/utils/annotation.go
xuezhaojun: If you are not using it either, I think we are safe to delete it.

pkg/utils/annotation.go
xuezhaojun: Yes, should I add a link to `addon-framework` to describe this key, and I saw `constant` pkg have this key, I don't whether it is ok to use it directly: https://github.com/open-cluster-management-io/addon-framework/blob/12c9395bde82925b8d238e300a0d9cae855f91d8/pkg/addonmanager/constants/constants.go#L46

xuezhaojun: Fixed.

pkg/utils/annotation.go
xuezhaojun: Removed.

pkg/addon/addon.go
```diff
@@ -45,8 +46,9 @@ type GlobalValues struct {
 }
 
 type Values struct {
-	Product      string       `json:"product,omitempty"`
-	GlobalValues GlobalValues `json:"global,omitempty,omitempty"`
+	Product                         string       `json:"product,omitempty"`
+	GlobalValues                    GlobalValues `json:"global,omitempty,omitempty"`
+	EnableSyncLabelsToClusterClaims bool         `json:"enableSyncLabelsToClusterClaims"`
```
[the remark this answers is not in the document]
xuezhaojun: No,we will add another flag for node collector in another PR.

PR #500, opened 2022-07-14
deprecate bm asset

ldpliu:
https://github.com/stolostron/backlog/issues/24149


zhiweiyin318:
do we also need to delete the inventory controller ?

PR #503, opened 2022-07-15
do not allow to update global clusterset

ldpliu:
https://github.com/stolostron/backlog/issues/24286

### DIFF
--- a/pkg/webhook/clusterset/validatingWebhook.go
+++ b/pkg/webhook/clusterset/validatingWebhook.go
@@ -81,13 +81,21 @@ func (a *AdmissionHandler) validateResource(request *v1.AdmissionRequest) *v1.Ad
 			}
 			return status
 		}
-		//allow create/update legacy clusterset
-		if clusterset.Spec.ClusterSelector.SelectorType == clusterv1beta1.LegacyClusterSetLabel {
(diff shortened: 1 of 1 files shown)


pkg/webhook/clusterset/validatingWebhook.go
```diff
@@ -81,13 +81,21 @@ func (a *AdmissionHandler) validateResource(request *v1.AdmissionRequest) *v1.Ad
 			}
 			return status
 		}
-		//allow create/update legacy clusterset
-		if clusterset.Spec.ClusterSelector.SelectorType == clusterv1beta1.LegacyClusterSetLabel {
+		//allow create/update global clusterset
+		if clusterset.Name == clustersetutils.GlobalSetName {
```
qiujian16, line 85: I think you have the validation on create, can we reuse the same logic?

ldpliu: > I think you have the validation on create, can we reuse the same logic?

The logic here is used for both creating and updating a set.

ldpliu: > if so, the message seems incorrect when create?

Update the message.

PR #509, opened 2022-07-18
Support hosted mode

zhujian7:
Signed-off-by: zhujian <jiazhu@redhat.com>
This PR will make the work manager addon support deploying agent in Hosted mode:
image

issue reference: https://github.com/stolostron/backlog/issues/23827

TODO:
- [ ] know issue: the length of the install namespace for hosted mode can not > 37, `63-len("klusterlet-addon-workmgr-")`(will doc this for SD)
```
- lastTransitionTime: "2022-07-18T04:14:38Z"
  message: 'Failed to apply manifest: Route.route.openshift.io "klusterlet-addon-workmgr"
    is invalid: spec.host: Invalid value: "klusterlet-addon-workmgr-klusterlet-hypershift-aws-demo-l7hlf-addon-test.apps.server-foundation-aws-410-cqv4j.dev04.red-chesterfield.com":
    must be no more than 63 characters'
  reason: AppliedManifestFailed
  status: "False"
  type: Applied
```
- [ ] the ca problem(will fix in a follow-up PR)
```
E0719 08:46:04.498234       1 nodecollecter.go:143] failed to get ca: namespaces "hypershift-aws-demo-l7hlf-addon-test" not found
E0719 08:46:40.521851       1 nodecollecter.go:143] failed to get ca: namespaces "hypershift-aws-demo-l7hlf-addon-test" not found
```
- [ ] integration test or e2e test(will add in a follow-up PR)


cmd/controller/app/server.go
```diff
@@ -58,6 +58,12 @@ var (
 	scheme = runtime.NewScheme()
 )
 
+const (
+	// deployModeAnnotationKey describe the klusterlet deploy mode when importing a managed cluster.
+	deployModeAnnotationKey string = "import.open-cluster-management.io/klusterlet-deploy-mode"
```
xuezhaojun, line 63: It's seems every addon need to define those vars.
Is there a way to make a consist pkg in addon-framework and addons can import and use?
[Not a blocker, it's ok in either way]

zhujian7: this annotation is only used in ACM, so I think there is not a good reason to put it in addon-framework.

zhujian7:
cc @ldpliu @qiujian16

PR #512, opened 2022-07-19
Change managedcluster name to uid of kube-system ns uid

xuezhaojun:

https://github.com/stolostron/backlog/issues/23826

### DIFF
--- a/pkg/klusterlet/clusterclaim/clusterclaimer.go
+++ b/pkg/klusterlet/clusterclaim/clusterclaimer.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	clusterv1beta1 "github.com/stolostron/cluster-lifecycle-api/clusterinfo/v1beta1"
+	"k8s.io/apimachinery/pkg/util/uuid"
 
 	configv1 "github.com/openshift/api/config/v1"
 	openshiftclientset "github.com/openshift/client-go/config/clientset/versioned"
@@ -108,7 +109,37 @@ type ClusterClaimer struct {
 	ConfigV1Client                  openshiftclientset.Interface
 	OauthV1Client                   openshiftoauthclientset.Interface
 	Mapper                          meta.RESTMapper
+	managedclusterID                string
 	EnableSyncLabelsToClusterClaims bool
+}
+
+// getManagedClusterID returns the managed cluster ID of the cluster.
+// We get ID by the sequence of: openshiftID, uid of kube-system namespace, random UID.
+func (c *ClusterClaimer) getManagedClusterID() (string, error) {
+	if c.managedclusterID != "" {
+		return c.managedclusterID, nil
+	}
+
+	isocp, err := c.isOpenShift()
+	if err != nil {
+		klog.Errorf("failed to check if the cluster is openshift.err:%v", err)
+		return "", err
+	}
+	if isocp {
+		_, ocpID, err := c.getOCPVersion()
+		if err == nil {
+			return ocpID, nil
+		}
+		klog.Errorf("Get ocpID failed, %v", err)
+	}
+
+	ns, err := c.KubeClient.CoreV1().Namespaces().Get(context.TODO(), "kube-system", metav1.GetOptions{})
+	if err == nil {
+		return string(ns.UID), nil
+	}
+	klog.Errorf("Get kube-system namespace failed, %v", err)
+
+	return string(uuid.NewUUID()), nil
 }
 
 func (c *ClusterClaimer) List() ([]*clusterv1alpha1.ClusterClaim, error) {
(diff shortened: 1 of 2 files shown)


qiujian16 (commented):
I think you can use a uuid directly. why do we need kub-system uid?

xuezhaojun:
1. use kube-system ns uid
2. if don't get enough permission, create a uuid as the value of id.k8s.io

pkg/klusterlet/clusterclaim/clusterclaimer.go
```diff
@@ -95,14 +97,48 @@ func init() {
...
+) (ClusterClaimer, error) {
+	var managedclusterID string
+	// If clusterclaim is not created, get UID of kube-system namespace
+	ns, err := kubeClient.CoreV1().Namespaces().Get(context.TODO(), "kube-system", metav1.GetOptions{})
+	if err != nil && !errors.IsForbidden(err) {
+		klog.Errorf("failed to get kube-system namespace. err:%v", err)
+		return ClusterClaimer{}, err
+	}
+	// Create a new uuid as the managedclusterID
+	if errors.IsForbidden(err) {
```
qiujian16, line 127: this should be openshift cluster id at first, then kube-system id, then random id.

xuezhaojun: Yes, we discussed the above process before, but I found we need to create clustercliamer first to get the openshiftID, let me see if I move these things before create.

qiujian16 (commented):
/approve

pkg/klusterlet/clusterclaim/clusterclaimer.go
```diff
@@ -199,8 +250,8 @@ func newClusterClaim(name, value string) *clusterv1alpha1.ClusterClaim {
 	}
 }
 
-func (c *ClusterClaimer) isOpenShift() (bool, error) {
-	_, err := c.Mapper.RESTMapping(schema.GroupKind{Group: "project.openshift.io", Kind: "Project"}, "v1")
+func isOpenShift(mapper meta.RESTMapper) (bool, error) {
```
zhujian7, line 253: why do we need to pass the mapper?

xuezhaojun: That is because we need to use this function before ClusterClaimer is created.

pkg/klusterlet/clusterclaim/clusterclaimer.go
```diff
@@ -95,14 +96,64 @@ func init() {
...
+		ConfigV1Client:   configV1Client,
+		OauthV1Client:    oauthV1Client,
+		Mapper:           mapper,
+		managedclusterID: managedClusterID,
+	}, nil
+}
+
+// getManagedClusterID returns the managed cluster ID of the cluster.
+// We get ID by the sequence of: openshiftID, uid of kube-system namespace, random UID.
+func getManagedClusterID(kubeClient kubernetes.Interface, mapper meta.RESTMapper, configV1Client openshiftclientset.Interface) (string, error) {
```
zhiweiyin318, line 136: I could define like this :
```
func (c *ClusterClaimer) getManagedClusterID() (string, error) {
   if c.managedclusterID != "" {
       return c. managedclusterID
   }
   .....
}
```
it will not change the  `c.isOpenShift` method.

xuezhaojun: Update, but that also means controller will send request to determine if this is an openshift every sync. Not sure whether we should take it into account.
Please take another look.

pkg/klusterlet/clusterclaim/clusterclaimer.go
```diff
@@ -95,14 +96,44 @@ func init() {
...
+	if err != nil {
+		klog.Errorf("failed to check if the cluster is openshift.err:%v", err)
+		return "", err
+	}
+	if isocp {
+		ocpID, err := getOCPID(c.ConfigV1Client)
+		if err == nil {
+			return ocpID, nil
+		}
+		klog.Errorf("Get ocpID failed, %v", err)
```
zhiweiyin318, line 127: as long as the managedclusterID has value, managedclusterID will not be changed.
should consider to reconcile the OCPID or namespaceID? it will be changed after work-manger restarts.

pkg/klusterlet/clusterclaim/clusterclaimer.go
```diff
@@ -255,6 +291,15 @@ const (
 	OCP3Version = "3"
 )
 
+func getOCPID(configv1Client openshiftclientset.Interface) (clusterID string, err error) {
```
zhiweiyin318, line 294: we already have a method to get clusterID `version, clusterID, err := c.getOCPVersion()` ,and the claim name is ClaimOpenshiftID.

pkg/klusterlet/clusterclaim/clusterclaimer.go
xuezhaojun: The `id.k8s.io` now is created only, means once it's created it would update.

pkg/klusterlet/clusterclaim/clusterclaimer.go
xuezhaojun: Yes,the design current is using openshiftID -> namespace uid -> randomID, but I think finally we should come up with a consist way to generate this ID.

PR #513, opened 2022-07-19
WIP: Add subfix to "label synced" clusterclaims name

xuezhaojun:
https://github.com/stolostron/backlog/issues/23826


xuezhaojun:
Using part of uuid means  once we have about 65536 id, it would cause collision, but I assume a managed cluster can not have such numbers of hub on it.

xuezhaojun:
/hold 
As adding prefix should be a long term thing, we need to clarify what "hubID" is and how we generate it.
In the short term, we add a flag to disable syncLabelsToClusterClaim function on SD hub.

PR #526, opened 2022-08-25
use restmapper to check ocp and do not delete version claim

zhiweiyin318:
issue: https://github.com/stolostron/backlog/issues/25185

### DIFF
--- a/pkg/klusterlet/clusterinfo/clusterinfo_controller.go
+++ b/pkg/klusterlet/clusterinfo/clusterinfo_controller.go
@@ -98,13 +98,8 @@ func (r *ClusterInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		r.isIbmCloud = true
 	}
 
-	// Get distribution info
-	newStatus.DistributionInfo, err = r.getDistributionInfo(ctx)
-	if err != nil {
-		log.Error(err, "Failed to get distribution info")
-		errs = append(errs, fmt.Errorf("failed to get distribution info, error:%v ", err))
-	}
-
+	product := ""
+	ocpVersion := ""
 	claims, err := r.ClaimLister.List(labels.Everything())
 	if err != nil {
 		log.Error(err, "Failed to list claims.")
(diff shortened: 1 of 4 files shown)


pkg/klusterlet/clusterinfo/clusterinfo_controller.go
```diff
@@ -329,18 +331,17 @@ func (r *ClusterInfoReconciler) getKubeVendor(product string) (kubeVendor cluste
 	return
 }
 
-func (r *ClusterInfoReconciler) isOpenshift() bool {
-	serverGroups, err := r.KubeClient.Discovery().ServerGroups()
+func (r *ClusterInfoReconciler) isOpenshift() (bool, error) {
```
zhujian7, line 334: Is it better to move this func to a common package, so it can be used in common?

zhiweiyin318: changed to check if OCP via product claim.

pkg/klusterlet/clusterinfo/clusterinfo_controller.go
```diff
@@ -121,6 +115,7 @@ func (r *ClusterInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			newStatus.ClusterID = value
 		case clusterclaim.ClaimOCMProduct:
 			newStatus.KubeVendor = r.getKubeVendor(value)
+			product = value
```
zhujian7, line 118: I am not quite sure using this claim is the best approach, I mean if there is a new supported product(ex: microshift maybe), we should add the new product type in https://github.com/stolostron/multicloud-operators-foundation/pull/526/files#diff-ac6674a4f427b422665e2fbea90ad4e776f2946ae22cc9974e4eb9045d5a4312R452 right? and I think it's a little bit error-prone(easy to forget to change).

Is there always a claim to indicate it is OCP no matter what the OCP product is? `id.openshift.io` or `version.openshift.io`?

pkg/klusterlet/clusterinfo/clusterinfo_controller.go
```diff
@@ -130,6 +125,13 @@ func (r *ClusterInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 	}
 
+	// Get distribution info
+	newStatus.DistributionInfo, err = r.getDistributionInfo(ctx, product)
```
zhujian7, line 129: Line 123 will be dead code. https://github.com/stolostron/multicloud-operators-foundation/pull/526/files#diff-ac6674a4f427b422665e2fbea90ad4e776f2946ae22cc9974e4eb9045d5a4312R123

zhujian7 (changes requested):

pkg/klusterlet/clusterinfo/clusterinfo_controller.go
```diff
@@ -124,12 +119,17 @@ func (r *ClusterInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 		case clusterclaim.ClaimOpenshiftVersion:
-			if newStatus.DistributionInfo.Type == clusterv1beta1.DistributionTypeOCP {
-				newStatus.DistributionInfo.OCP.Version = value
-			}
+			ocpVersion = value
 		}
 	}
 
+	// Get distribution info
+	newStatus.DistributionInfo, err = r.getDistributionInfo(ctx, product, ocpVersion)
```
zhujian7, line 127: the product will always be empty?

zhujian7 (changes requested):

PR #549, opened 2022-12-15
Fix: auto-detect cloud and vendor labels when label are not set or eq…

xuezhaojun:
…uals to empty.

Fixes: https://issues.redhat.com/browse/ACM-2375


xuezhaojun:
/hold

This is an api-level change, it will cause customers can not delete those 2 labels.

PR #552, opened 2023-01-20
update id.k8s.io claim if empty

zhiweiyin318:

issue :  https://issues.redhat.com/browse/ACM-2998

### DIFF
--- a/pkg/klusterlet/clusterclaim/clusterclaim_controller.go
+++ b/pkg/klusterlet/clusterclaim/clusterclaim_controller.go
@@ -97,7 +97,7 @@ func (r *ClusterClaimReconciler) createOrUpdate(ctx context.Context, newClaim *c
 		return fmt.Errorf("unable to get ClusterClaim %q: %w", newClaim.Name, err)
 	case !reflect.DeepEqual(oldClaim.Spec, newClaim.Spec):
 		// if newClaim.Name is in clusterClaimCreateOnlyList, then do nothing
-		if utils.ContainsString(clusterClaimCreateOnlyList, newClaim.Name) {
+		if utils.ContainsString(clusterClaimCreateOnlyList, newClaim.Name) && oldClaim.Spec.Value != "" {
 			return nil
 		}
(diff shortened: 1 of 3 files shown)


pkg/klusterlet/clusterclaim/clusterclaim_controller.go
```diff
@@ -97,7 +97,7 @@ func (r *ClusterClaimReconciler) createOrUpdate(ctx context.Context, newClaim *c
 		return fmt.Errorf("unable to get ClusterClaim %q: %w", newClaim.Name, err)
 	case !reflect.DeepEqual(oldClaim.Spec, newClaim.Spec):
 		// if newClaim.Name is in clusterClaimCreateOnlyList, then do nothing
-		if utils.ContainsString(clusterClaimCreateOnlyList, newClaim.Name) {
+		if utils.ContainsString(clusterClaimCreateOnlyList, newClaim.Name) && oldClaim.Spec.Value != "" {
```
qiujian16, line 100: do you need to also ensure newClaim is not empty?

zhiweiyin318: check in https://github.com/stolostron/multicloud-operators-foundation/pull/552/files#diff-c78e2a145aa414b25e54617166eced7480b482f0683ae9736ece7497acdbe2d5R175 
if clusterID is empty, will not newClaim

zhiweiyin318: added a check in syncClaims before call createOrUpdate

zhiweiyin318: @elgnay  do we have empty value claim scenario? I found there are empty value claim case in the unit tests.

zhiweiyin318: In this case, for the clam in clusterClaimCreateOnlyList , do not need to update when the old claim has valid value.

zhiweiyin318:
cherry-pick /backplane-2.2

PR #555, opened 2023-02-01
Fix: klog log verbosity.

xuezhaojun:


xuezhaojun:
As in https://github.com/kubernetes/klog/blob/main/README.md#how-to-use-klog says, 
```
Use klog.InitFlags(nil) explicitly for initializing global flags as we no longer use init() method to register the flags
```

PR #560, opened 2023-02-14
update the criteria for determining worker nodes

zhiweiyin318:
core_worker and socket_worker are collected by acm_managed_cluster_info metric for billing, 
so need calculate all nodes where the workloads can be placed. so calculate the data from all nodes we can get .

story: https://issues.redhat.com/browse/ACM-3464

### DIFF
--- a/pkg/controllers/clusterinfo/capacity_controller.go
+++ b/pkg/controllers/clusterinfo/capacity_controller.go
@@ -17,9 +17,12 @@ import (
 )
 
 const (
-	resourceSocket       clusterv1.ResourceName = "socket"
-	resourceCoreWorker   clusterv1.ResourceName = "core_worker"
-	resourceSocketWorker clusterv1.ResourceName = "socket_worker"
+	resourceSocket               clusterv1.ResourceName = "socket"
+	resourceCoreWorker           clusterv1.ResourceName = "core_worker"
+	resourceSocketWorker         clusterv1.ResourceName = "socket_worker"
+	LabelNodeRoleOldControlPlane                        = "node-role.kubernetes.io/master"
+	LabelNodeRoleControlPlane                           = "node-role.kubernetes.io/control-plane"
+	LabelNodeRoleInfra                                  = "node-role.kubernetes.io/infra"
 )
 
 type CapacityReconciler struct {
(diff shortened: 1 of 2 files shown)


pkg/controllers/clusterinfo/capacity_controller.go
```diff
@@ -62,11 +62,12 @@ func (r *CapacityReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+
+	// core_worker and socket_worker are used for billing.
+	// so need calculate all nodes where workload can be placed.
 	for _, node := range nodes {
-		if isWorker(node) {
-			coreWorkerCapacity.Add(node.Capacity[clusterv1.ResourceCPU])
-			socketWorkerCapacity.Add(node.Capacity[resourceSocket])
-		}
+		coreWorkerCapacity.Add(node.Capacity[clusterv1.ResourceCPU])
+		socketWorkerCapacity.Add(node.Capacity[resourceSocket])
```
qiujian16, line 70: I remember we do not count master and infra node before, does that change?

zhiweiyin318: discussed with @jeff-brent  in the doc https://docs.google.com/document/d/1E78EKIB5TJ52u2RHeHLbG0O_h5DSu2W_OMs9aYfrfSM/edit?usp=sharing
we need to count all nodes where workloads can be placed for billing. the user can deploy workloads on the master/infra nodes if set the tolerations. so it is inaccurate to use master/infra label to determine which are worker nodes. I think all nodes we can get are all worker nodes.

qiujian16: then why we still need coreworker? it is same as capacity that we have collected in registration

zhiweiyin318: I misunderstood the openshift subscription billing rules. The master and infra nodes are not included in the subscription cost calculation. and in OSD and ROSA it's not possible to create MachineSets/MachineConfigPools. Also, the infra nodes in OSD and ROSA are managed by Red Hat, and customer workloads cannot be executed there.
so the core calculation is change to :
1. for OCP, omit the master/infra nodes.
2. for non-OCP, calculate all nodes.

PR #566, opened 2023-02-20
add a webhook to prevent the hosting cluster from being detached mistakenly

zhujian7:


zhujian7:
cc @rokej

pkg/constants/constants.go
```diff
@@ -0,0 +1,15 @@
+package constants
+
+const (
+	// KlusterletDeployModeAnnotation is the annotation key of klusterlet deploy mode, it describes the
+	// klusterlet deploy mode when importing a managed cluster.
+	// If the value is "Hosted", the HostingClusterNameAnnotation annotation will be required, we use
+	// AnnotationKlusterletHostingClusterName to determine where to deploy the registration-agent and
+	// work-agent.
+	AnnotationKlusterletDeployMode string = "import.open-cluster-management.io/klusterlet-deploy-mode"
```
zhiweiyin318, line 9: suggest ot move the constants to clc-api repo https://github.com/stolostron/cluster-lifecycle-api .
it could be more easy to manage and use across the repos.

zhujian7:
@zhiweiyin318 PTAL

pkg/webhook/validating/webhook.go
```diff
@@ -198,6 +202,29 @@ func (a *AdmissionHandler) validateManagedCluster(request *v1.AdmissionRequest)
...
+			if cluster.GetAnnotations()[constants.AnnotationKlusterletDeployMode] != "Hosted" {
+				continue
+			}
+			if cluster.GetAnnotations()[constants.AnnotationKlusterletHostingClusterName] != clusterName {
+				continue
+			}
+			hostedClusters = append(hostedClusters, cluster.Name)
+		}
+
+		if len(hostedClusters) > 0 {
```
rokej, line 223: @zhujian7 I don't understand how this works. If there are multiple hosting clusters, how do you get the list of hosted clusters for a specific hosting cluster that the user chooses to delete?

zhujian7: there is a condition check `cluster.GetAnnotations()[constants.AnnotationKlusterletHostingClusterName] != clusterName` on line 217.

rokej: Ah OK Thanks @zhujian7

PR #567, opened 2023-02-20
fix kube informer bug

ldpliu:


cmd/controller/app/server.go
```diff
@@ -278,6 +283,18 @@ func Run(o *options.ControllerRunOptions, ctx context.Context) error {
 	go func() {
```
zhujian7, line 283: is this go routine necessary?

cmd/controller/app/server.go
```diff
@@ -278,6 +289,18 @@ func Run(o *options.ControllerRunOptions, ctx context.Context) error {
...
 		<-mgr.Elected()
 
+		kubeInfomers.Start(ctx.Done())
+		clusterInformers.Start(ctx.Done())
+
+		if ok := kubecache.WaitForCacheSync(ctx.Done(),
```
qiujian16, line 295: this wait does not seems to be necessary

cmd/controller/app/server.go
```diff
@@ -113,6 +115,15 @@ func Run(o *options.ControllerRunOptions, ctx context.Context) error {
 	clusterInformers := clusterv1informers.NewSharedInformerFactory(clusterClient, 10*time.Minute)
 	kubeInfomers := kubeinformers.NewSharedInformerFactory(kubeClient, 10*time.Minute)
 
+	clusterRoleBindingsInformer := kubeInfomers.Rbac().V1().ClusterRoleBindings()
+	clusterRolesInformer := kubeInfomers.Rbac().V1().ClusterRoles()
+	roleBindingsInformer := kubeInfomers.Rbac().V1().RoleBindings()
+
+	//Init the used kube Informer, if we do not call it expicity, the related informer may not run.
+	clusterRoleBindingsInformer.Informer()
```
qiujian16, line 123: I do not fully understand, you have called this in the sync controller.

pkg/controllers/clusterset/syncclusterrolebinding/syncclusterrolebinding_controller.go
```diff
@@ -50,10 +48,6 @@ func NewReconciler(kubeClient kubernetes.Interface,
 
 // Run a routine to sync the clusterrolebinding periodically.
 func (r *Reconciler) Run(ctx context.Context, period time.Duration) {
-	if ok := kubecache.WaitForCacheSync(ctx.Done(), r.clusterRolebindingInformer.Informer().HasSynced); !ok {
```
qiujian16, line 53: I'd rather having r.clusterRolebindingInformer.Informer() set during NewReconciler which then gives you a hassynced func.

cmd/controller/app/server.go
ldpliu: I move the waitforcachesync func outside the sync controller. As I want to just use the lister interface in sync controller.

cmd/controller/app/server.go
ldpliu: Previously, we only call the `rolesBindingInformer.Informer().HasSynced` in waitforcachesync func, which is after the kube informer start. 
So when the kubeinformer start, the rolebinding informer will not start.

pkg/controllers/clusterset/syncclusterrolebinding/syncclusterrolebinding_controller.go
ldpliu: As @zhujian7 suggested, I use the lister interface in newReconciler, I think it's better.

pkg/controllers/clusterset/syncclusterrolebinding/syncclusterrolebinding_controller.go
```diff
@@ -50,10 +49,6 @@ func NewReconciler(kubeClient kubernetes.Interface,
...
-		klog.Errorf("failed to wait for kubernetes caches to sync")
-		return
-	}
 	go utilwait.Forever(r.reconcile, period)
```
zhujian7, line 52: the outside caller is invoking this func as `go Run()`, so I think the `go` keyword is not necessary, BTW, I think utilwait.UntilWithContext is better than utilwait.Forever

zhujian7: and please check others `Run` from other packages.

cmd/controller/app/server.go
```diff
@@ -278,6 +283,18 @@ func Run(o *options.ControllerRunOptions, ctx context.Context) error {
 	go func() {
 		<-mgr.Elected()
 
+		kubeInfomers.Start(ctx.Done())
```
zhujian7, line 286: consider moving the informer.Start outside the goroutine, because when the mgr stars running, it needs the informer, right? so putting it ahead of `mgr.Start` is better in my opinion

pkg/controllers/clusterset/syncclusterrolebinding/syncclusterrolebinding_controller.go
```diff
@@ -50,15 +49,10 @@ func NewReconciler(kubeClient kubernetes.Interface,
 
 // Run a routine to sync the clusterrolebinding periodically.
```
zhujian7, line 50: correct the comments?

pkg/controllers/clusterset/syncrolebinding/syncrolebinding_controller.go
```diff
@@ -53,18 +52,11 @@ func NewReconciler(kubeClient kubernetes.Interface,
 
 // start a routine to sync the clusterrolebinding periodically.
 func (r *Reconciler) Run(ctx context.Context, period time.Duration) {
```
zhujian7, line 54: ditto

pkg/controllers/clusterset/syncclusterrolebinding/syncclusterrolebinding_controller.go
zhujian7: please check the `pkg/controllers/gc/garbagecollector_controller.go Run()` as well. no need go routine.

qiujian16 (commented):
/lgtm

PR #568, opened 2023-02-21
fix kube informer bug to mce 2.2

ldpliu:
manually- cherry-pick https://github.com/stolostron/multicloud-operators-foundation/pull/567


ldpliu:
/hold
it looks 2.7.1 code freeze, so need z stream open

PR #571, opened 2023-03-02
fix ocp version

ldpliu:


pkg/klusterlet/clusterinfo/distributionInfo_sync.go
```diff
@@ -77,7 +79,10 @@ func (s *distributionInfoSyncer) syncOCPDistributionInfo(ctx context.Context, cl
 		history.Verified = historyItem.Verified
 		ocpDistributionInfo.VersionHistory = append(ocpDistributionInfo.VersionHistory, history)
 		if historyItem.State == "Completed" {
```
zhiweiyin318, line 81: could we get ocp version from clusterClaim here?

pkg/klusterlet/clusterclaim/clusterclaimer.go
```diff
@@ -338,10 +338,14 @@ func (c *ClusterClaimer) getOCPVersion() (version, clusterID string, err error)
...
 	historyItems := clusterVersion.Status.History
+	var latestCompleteTime *metav1.Time
+
 	for _, historyItem := range historyItems {
 		if historyItem.State == "Completed" {
-			version = historyItem.Version
-			break
+			if latestCompleteTime.IsZero() || latestCompleteTime.Before(historyItem.CompletionTime) {
+				latestCompleteTime = historyItem.CompletionTime
+				version = historyItem.Version
```
zhujian7, line 347: Could we rename the v`ersion` to `latestCompleteVersion` to match the `latestCompleteTime`

pkg/klusterlet/clusterinfo/distributionInfo_sync.go
```diff
@@ -76,10 +79,15 @@ func (s *distributionInfoSyncer) syncOCPDistributionInfo(ctx context.Context, cl
...
 		history.Verified = historyItem.Verified
 		ocpDistributionInfo.VersionHistory = append(ocpDistributionInfo.VersionHistory, history)
-		if historyItem.State == "Completed" {
-			ocpDistributionInfo.Version = historyItem.Version
-		}
 	}
+
+	version, err := s.claimLister.Get(clusterclaim.ClaimOpenshiftVersion)
+	if err != nil {
+		klog.Errorf("Failed to get ocp version, error: %v", err)
```
zhujian7, line 86: should we return the err?

pkg/klusterlet/clusterclaim/clusterclaimer.go
ldpliu: done

pkg/klusterlet/clusterinfo/distributionInfo_sync.go
ldpliu: > should we return the err?

I do not think it's necessary to return err. As other fields should also be updated.

zhujian7: think about when will return an error here? if we can not get claim(except for the not found error), can other places get the claim successfully?
IMO we should return the error if it is not a `NotFound` error.

zhujian7:
LGTM

PR #577, opened 2023-03-14
set leader election options

ldpliu:


cmd/controller/app/options/options.go
```diff
@@ -49,6 +55,12 @@ func (o *ControllerRunOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&o.EnableLeaderElection, "enable-leader-election", o.EnableLeaderElection,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	fs.IntVar(&o.LeaseDuration, "lease-duration", o.LeaseDuration,
```
xuezhaojun, line 58: Can we use `fs.DurationVar` instead of `fs.IntVar`, it will save the convert process.

```
fs.DurationVar(&o.LeaseDuration, "lease-duration", o.LeaseDuration,
	"Non-leader candidates will wait to force acquire leadership. Default is 15 seconds.")
```

ldpliu: > Can we use `fs.DurationVar` instead of `fs.IntVar`, it will save the convert process.
> 
> ```
> fs.DurationVar(&o.LeaseDuration, "lease-duration", o.LeaseDuration,
> 	"Non-leader candidates will wait to force acquire leadership. Default is 15 seconds.")
> ```

Thank you.
I did not notice this func previously, changed.

xuezhaojun: NP！ looks good to me now👍

PR #578, opened 2023-03-14
Remove: webhook admission of annotation(not in used anymore)

xuezhaojun:


xuezhaojun:
/close

Since the placementrule is still in used, we should also keep this feature at the point of time.

PR #587, opened 2023-03-29
Remove rbac

ldpliu:


cmd/controller/app/server.go
```diff
@@ -57,7 +57,8 @@ import (
 )
 
 var (
-	scheme = runtime.NewScheme()
+	scheme      = runtime.NewScheme()
+	disableRBAC = true
```
qiujian16, line 61: we need this to be a flag I think

PR #592, opened 2023-03-30
do not create route in hosted mode

zhiweiyin318:
issue: https://issues.redhat.com/browse/ACM-4652


pkg/klusterlet/clusterinfo/controller.go
```diff
@@ -84,7 +85,7 @@ func (r *ClusterInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			claimLister:          r.ClaimLister,
 		},
 	}
-	if !r.DisableLoggingInfoSyncer {
+	if !r.DisableLoggingInfoSyncer || r.InstallMode != "Hosted" {
```
qiujian16, line 88: curious: when do we set DisableLoggingInfoSyncer to true?

zhiweiyin318: the controller is imported in  standalone controlplane, and started with DisableLoggingInfoSyncer == true.

zhiweiyin318: DisableLoggingInfoSyncer is not a flag of agent binary, let me use it to replace InstallMode.

PR #594, opened 2023-04-12
add seccomprofile in security context

ldpliu:


pkg/addon/manifests/chart/templates/deployment.yaml
```diff
@@ -21,6 +21,12 @@ spec:
         component: "work-manager"
     spec:
       serviceAccountName: {{ template "work-manager.fullname" . }}
+      securityContext:
+      {{- if ne .Values.installMode "Hosted" }}
+        seccompProfile:
+          type: RuntimeDefault 
+      {{- end }}
+        runAsNonRoot: true
```
zhiweiyin318, line 29: do we need this? it has been configured in containers.

PR #605, opened 2023-05-10
replace the last applied managed cluster api server url when it changes

zhujian7:
issue reference: https://issues.redhat.com/browse/ACM-4647

### DIFF
--- a/pkg/controllers/clusterca/clusterca_controller.go
+++ b/pkg/controllers/clusterca/clusterca_controller.go
@@ -7,9 +7,9 @@ import (
 	clusterinfov1beta1 "github.com/stolostron/cluster-lifecycle-api/clusterinfo/v1beta1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog"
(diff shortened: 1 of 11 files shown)


pkg/controllers/clusterca/clusterca_controller.go
```diff
@@ -56,50 +56,82 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
-			return clusterConfigs, true
+
+		if reflect.DeepEqual(cluConfig.CABundle, clusterinfoConfig.CABundle) && cluConfig.URL == clusterinfoConfig.URL {
+			return clusterConfigs, false, false
 		}
-		return clusterConfigs, false
+
+		clusterConfigs[i].URL = clusterinfoConfig.URL
+		clusterConfigs[i].CABundle = clusterinfoConfig.CABundle
+		return clusterConfigs, true, cluConfig.URL != clusterinfoConfig.URL
```
qiujian16, line 134: cluConfig.URL != clusterinfoConfig.URL this seems buggy, what happens if updateMC succeed and updateMCI fails? I think you should always check if lastApplied equals clusterinfoConfig after updateMC succeeds

zhujian7: fixed, will always check if lastApplied equals clusterinfoConfig no matter  managedCluster needs to update or not.

PR #657, opened 2023-09-11
Fix: missing clusterclaim.

xuezhaojun:
Fixes: https://issues.redhat.com/browse/ACM-7408

### DIFF
--- a/pkg/klusterlet/clusterclaim/clusterclaimer.go
+++ b/pkg/klusterlet/clusterclaim/clusterclaimer.go
@@ -101,7 +101,7 @@ var ProductOCPList = []string{
 }
 
 // internalLabels includes the labels managed by ACM.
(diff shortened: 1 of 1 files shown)


pkg/klusterlet/clusterclaim/clusterclaimer.go
```diff
@@ -101,16 +101,14 @@ var ProductOCPList = []string{
 }
 
 // internalLabels includes the labels managed by ACM.
-var internalLabels = sets.String{}
+var internalLabels = sets.Set[string]{}
```
xuezhaojun, line 104:
```go
// Deprecated: use generic Set instead.
// new ways:
// s1 := Set[string]{}
// s2 := New[string]()
type String map[string]Empty
```

Update the depracated function.

PR #659, opened 2023-09-13
Fix: return origin error when it mapping fails.

xuezhaojun:
fixes: https://issues.redhat.com/browse/ACM-7479

### DIFF
--- a/pkg/utils/rest/kubecontrol.go
+++ b/pkg/utils/rest/kubecontrol.go
@@ -68,6 +68,7 @@ func NewKubeControl(mapper meta.RESTMapper, config *rest.Config) *KubeControl {
 // The resourceOrKindArg should be in the format of {resource}.{version}.{group}, or {resource}.
 // If it is {resource}.{group}, it will be mistakenly parsed to {resource: {resource}, version: {group or part of group if there is "." in group} }
 // which will trigger the reload of restmapper and introduce performance degradation.
+// This is a copy from: https://github.com/kubernetes/cli-runtime/blob/e7b1ca8f27e99b474a841c85a379bf25702dfcb9/pkg/resource/builder.go#L768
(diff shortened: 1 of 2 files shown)


pkg/utils/rest/kubecontrol.go
```diff
@@ -96,10 +97,7 @@ func MappingFor(mapper meta.RESTMapper, resourceOrKindArg string) (*meta.RESTMap
 
 	mapping, err := mapper.RESTMapping(groupKind, gvk.Version)
 	if err != nil {
-		// if we error out here, it is because we could not match a resource or a kind
-		// for the given argument. To maintain consistency with previous behavior,
-		// announce that a resource type could not be found.
-		return nil, fmt.Errorf("the server doesn't have a resource type %q", groupResource.Resource)
+		return nil, fmt.Errorf("fail to mapping GroupKind %v, GroupKindVersion %v, resource:%s err: %v", groupKind, gvk.Version, resourceOrKindArg, err)
```
xuezhaojun, line 100: This piece of code is copied from:https://github.com/kubernetes/cli-runtime/blob/e7b1ca8f27e99b474a841c85a379bf25702dfcb9/pkg/resource/builder.go#L768

But we need to return origin error instead of replacing to "type not found"

xuezhaojun:
agent is actually using `NewDynamicRESTMapper` in `apiutil` pkg of `controller-runtime`: https://github.com/stolostron/multicloud-operators-foundation/blob/aefb7784533343e7d760e8cd1d1859669719be49/cmd/agent/agent.go#L158

But u-t is using `NewDynamicRESTMapper` in `restmapper` pkg of `client-go`:
https://github.com/stolostron/multicloud-operators-foundation/blob/aefb7784533343e7d760e8cd1d1859669719be49/pkg/klusterlet/view/view_controller_test.go#L418

So it can not promise agent works.

PR #661, opened 2023-09-17
Fix: restmapping fail when new crd added.

xuezhaojun:
fixes: https://issues.redhat.com/browse/ACM-7479

`DefferdDiscoveryRestMapper` requires explicily reload when`IsNotMatch` error occurs.


qiujian16 (commented):
/approve
/lgtm

PR #662, opened 2023-09-21
Refactor: klusterlet claims

xuezhaojun:
Fixes: 
* https://issues.redhat.com/browse/ACM-7497: The agent on the hosting cluster continuously deletes the synced label claims that created by the customer hub.
* https://issues.redhat.com/browse/ACM-7500:  When rosa-logo configmap is created after agent pods started, it can't be detected as rosa.

### DIFF
--- a/cmd/agent/agent.go
+++ b/cmd/agent/agent.go
@@ -250,20 +250,24 @@ func startManager(o *options.AgentOptions, ctx context.Context) {
 			Agent:                    agent,
 		}
 		clusterClaimer := clusterclaimctl.ClusterClaimer{
-			ClusterName:                     o.ClusterName,
-			HubClient:                       mgr.GetClient(),
-			KubeClient:                      managedClusterKubeClient,
-			ConfigV1Client:                  openshiftClient,
-			OauthV1Client:                   osOauthClient,
-			Mapper:                          reloadMapper,
-			EnableSyncLabelsToClusterClaims: o.EnableSyncLabelsToClusterClaims,
+			KubeClient:     managedClusterKubeClient,
+			ConfigV1Client: openshiftClient,
+			OauthV1Client:  osOauthClient,
+			Mapper:         reloadMapper,
 		}
 
-		clusterClaimReconciler := clusterclaimctl.ClusterClaimReconciler{
-			Log:               ctrl.Log.WithName("controllers").WithName("ManagedClusterInfo"),
-			ClusterClient:     managedClusterClusterClient,
-			ClusterInformers:  clusterInformerFactory.Cluster().V1alpha1().ClusterClaims(),
-			ListClusterClaims: clusterClaimer.List,
+		clusterClaimReconciler, err := clusterclaimctl.NewClusterClaimReconciler(
+			ctrl.Log.WithName("controllers").WithName("ClusterClaim"),
+			o.ClusterName,
+			managedClusterClusterClient,
+			mgr.GetClient(),
+			clusterInformerFactory.Cluster().V1alpha1().ClusterClaims().Lister(),
+			clusterClaimer.GenerateExpectClusterClaims,
+			o.EnableSyncLabelsToClusterClaims,
+		)
+		if err != nil {
+			setupLog.Error(err, "unable to create controller", "controller", "ClusterClaim")
+			os.Exit(1)
 		}
 
 		if err = actionReconciler.SetupWithManager(mgr); err != nil {
@@ -281,7 +285,8 @@ func startManager(o *options.AgentOptions, ctx context.Context) {
 			os.Exit(1)
 		}
 
-		if err = clusterClaimReconciler.SetupWithManager(mgr); err != nil {
+		if err = clusterClaimReconciler.SetupWithManager(mgr,
+			clusterInformerFactory.Cluster().V1alpha1().ClusterClaims()); err != nil {
 			setupLog.Error(err, "unable to create controller", "controller", "ClusterClaim")
 			os.Exit(1)
 		}
--- a/pkg/klusterlet/clusterclaim/clusterclaim_controller.go
+++ b/pkg/klusterlet/clusterclaim/clusterclaim_controller.go
@@ -4,98 +4,295 @@ import (
 	"context"
 	"fmt"
 	"reflect"
-
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	"strings"
 
 	"github.com/go-logr/logr"
+	clusterv1beta1 "github.com/stolostron/cluster-lifecycle-api/clusterinfo/v1beta1"
 	"github.com/stolostron/multicloud-operators-foundation/pkg/utils"
 	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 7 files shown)


pkg/klusterlet/clusterclaim/clusterclaimer.go
```diff
@@ -158,16 +157,15 @@ func (c *ClusterClaimer) getManagedClusterID() (string, error) {
 	return string(uuid.NewUUID()), nil
 }
 
-func (c *ClusterClaimer) List() ([]*clusterv1alpha1.ClusterClaim, error) {
+func (c *ClusterClaimer) GenerateExpectClusterClaims() ([]*clusterv1alpha1.ClusterClaim, error) {
```
xuezhaojun, line 160: It is often confused with the common `List` operations by clusterClient when reading. Rename to `Generate`

pkg/klusterlet/clusterinfo/manager.go
```diff
@@ -20,7 +21,7 @@ func (r *ClusterInfoReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	claimSource := clusterclaim.NewClusterClaimSource(r.ClaimInformer)
 	nodeSource := &nodeSource{nodeInformer: r.NodeInformer.Informer()}
 	return ctrl.NewControllerManagedBy(mgr).
-		WatchesRawSource(claimSource, &clusterclaim.ClusterClaimEventHandler{}).
+		WatchesRawSource(claimSource, clusterclaim.NewClusterClaimEventHandler(labels.Everything())).
```
xuezhaojun, line 24: clusterinfo controller should consider all kinds of claims(including customer created claims that has no label)

pkg/klusterlet/clusterclaim/manager.go
```diff
@@ -55,13 +72,20 @@ func (s *clusterClaimSource) Start(ctx context.Context, handler handler.EventHan
 	// all predicates are ignored
 	s.claimInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
-			handler.Create(ctx, event.CreateEvent{}, queue)
+			handler.Create(ctx, event.CreateEvent{
```
xuezhaojun, line 75: We need to pass labels into handler now.

cmd/agent/agent.go
```diff
@@ -250,20 +250,33 @@ func startManager(o *options.AgentOptions, ctx context.Context) {
...
+			managedClusterClusterClient,
+			clusterInformerFactory.Cluster().V1alpha1().ClusterClaims().Lister(),
+			clusterClaimer.GenerateExpectClusterClaims,
+		)
+		if err != nil {
+			setupLog.Error(err, "unable to create controller", "controller", "ClusterClaim")
+			os.Exit(1)
+		}
+
+		labelClusterClaimReconciler, err := clusterclaimctl.NewLableClusterClaimReconicler(
```
zhiweiyin318, line 270: I don't think we need create another reconciler controller, we can use different sync functions in one controller like the common implementation  in ocm repo.

pkg/klusterlet/clusterclaim/labelclusterclaim_controller.go
```diff
@@ -0,0 +1,139 @@
...
+		if apierrors.IsNotFound(err) {
+			return claims, nil
+		}
+		return claims, err
+	}
+
+	// do not create claim if the label is managed by ACM.
+	// if the label format is aaa/bbb, the name of claim will be bbb.aaa.
+	// Besides, "_" and "/" in the label name will be replaced with "-" and "." respectively.
+	for label, value := range clusterInfo.Labels {
```
zhiweiyin318, line 103: we need sync the labels of managedCluster not managedClusterInfo, although the labels of managedcluster are included in the labels of managedClusterInfo.
but if meet error, the cluster labels are not synced to managedClusterInfo?

xuezhaojun: This function is a copy from the old code with no change: https://github.com/stolostron/multicloud-operators-foundation/blob/111059e6a651503d3b6c89e9c62cfe65be5718f9/pkg/klusterlet/clusterclaim/clusterclaimer.go#L715

Agree to change to sync from managedcluster,  how about leaving a todo and fixing this with another refactor PR?

cmd/agent/agent.go
xuezhaojun: Thanks for the advise, do we means multiple `syn`c in one controller here? But I didn't find the practice in the `ocm` repo, would you mind help me with an example?

Also worth to mention, a motivation of separating a new controller is using different eventhandlers to filter clusterclaim events. For example, if a `spoke-only` clusterclaim updates, we only want to trigger the reconcile of `labelClusterClaimReconciler`.

xuezhaojun: Thanks @zhiweiyin318 for providing the example: https://github.com/open-cluster-management-io/ocm/blob/5405099cb5b8e49a547f8e3771f65087fd7c0bfd/pkg/addon/controllers/addonmanagement/controller.go#L105

This is a good practice of simply the code, @qiujian16 would you mind taking a second look? I think as zhiwei said, this could also be a way of 'sperate controller'?

xuezhaojun: Confirmed with @qiujian16, we have no preference on this, both ways are ok, I will change to use one controller to follow the same style with other codes.

pkg/klusterlet/clusterclaim/clusterclaim_controller.go
```diff
@@ -5,97 +5,168 @@ import (
...
+// then reflect on the console.
+var updateExceptions = []updateException{
+	func(oldClaim, newClaim *clusterv1alpha1.ClusterClaim) bool {
+		return utils.ContainsString(clusterClaimCreateOnlyList, newClaim.Name) &&
+			oldClaim.Spec.Value != ""
+	},
+	func(oldClaim, newClaim *clusterv1alpha1.ClusterClaim) bool {
+		return oldClaim.Name == ClaimOCMProduct &&
+			oldClaim.Spec.Value != ProductOther &&
+			newClaim.Spec.Value == ProductOther
```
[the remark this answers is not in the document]
xuezhaojun: Fixed.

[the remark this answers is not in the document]
xuezhaojun: Fixed.

qiujian16 (commented):
/approve

pkg/klusterlet/clusterclaim/clusterclaim_controller.go
```diff
@@ -4,98 +4,303 @@ import (
...
+		// only check ClaimOCMProduct
+		if oldClaim.Name != ClaimOCMPlatform {
+			return true
+		}
+		// don't allow update from a specific value to "Other"
+		if newClaim.Spec.Value == PlatformOther {
+			return false
+		}
+		// do not allow update unless from other
+		if oldClaim.Spec.Value != PlatformOther {
```
xuezhaojun, line 252: @qiujian16 @zhiweiyin318 I think we should delete this because, in the rosa-host case, the configmap of rosa-logo has the chance of being created after the pod runs for a while. 

It will determine product as OCP first, but we want it turns to ROSA later.

PR #669, opened 2023-10-13
Fix: distributionInfo miss sync for OSD type.

xuezhaojun:
Fixes: https://issues.redhat.com/browse/ACM-7888

Also need to backport to 2.8.3.

### DIFF
--- a/pkg/klusterlet/clusterinfo/distributionInfo_sync.go
+++ b/pkg/klusterlet/clusterinfo/distributionInfo_sync.go
@@ -25,7 +25,7 @@ type distributionInfoSyncer struct {
 func (s *distributionInfoSyncer) sync(ctx context.Context, clusterInfo *clusterv1beta1.ManagedClusterInfo) error {
 	// currently we only support OCP in DistributionInfo
 	switch clusterInfo.Status.KubeVendor {
-	case clusterv1beta1.KubeVendorOpenShift:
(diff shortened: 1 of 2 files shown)


pkg/klusterlet/clusterinfo/distributionInfo_sync.go
```diff
@@ -25,7 +25,7 @@ type distributionInfoSyncer struct {
 func (s *distributionInfoSyncer) sync(ctx context.Context, clusterInfo *clusterv1beta1.ManagedClusterInfo) error {
 	// currently we only support OCP in DistributionInfo
 	switch clusterInfo.Status.KubeVendor {
-	case clusterv1beta1.KubeVendorOpenShift:
+	case clusterv1beta1.KubeVendorOpenShift, clusterv1beta1.KubeVendorOSD:
```
zhiweiyin318, line 28: not only OCP and OSD, but also ARO, ROSA, ROKS which are in OCPList

xuezhaojun: That's  a fun truth, in the `ProductOCPList`, other type will finally be represent by `KubeVendorOpenShift`: https://github.com/stolostron/multicloud-operators-foundation/blob/b4de56dec0b68530ad4701805403e2b66272e484/pkg/klusterlet/clusterinfo/defaultInfo_sync.go#L88

Issue #680, opened 2023-11-10
e2e: no secret found for service account in ocp 4.14


zhujian7:
Running e2e tests on an ocp 4.14 cluster, the e2e failed with:

```
Testing ManagedCluster Testing Clusterca sync Get CA from service account
/go/src/github.com/stolostron/multicloud-operators-foundation/test/e2e/managedcluster_test.go:158
  [FAILED] in [It] - /go/src/github.com/stolostron/multicloud-operators-foundation/test/e2e/managedcluster_test.go:164 @ 11/10/23 02:43:19.359
• [FAILED] [0.020 seconds]
Testing ManagedCluster Testing Clusterca sync [It] Get CA from service account
/go/src/github.com/stolostron/multicloud-operators-foundation/test/e2e/managedcluster_test.go:158
  [FAILED] Unexpected error:
      <*errors.errorString | 0xc000b338f0>: 
      secret with type kubernetes.io/service-account-token not found in service account kube-system/default
      {
          s: "secret with type kubernetes.io/service-account-token not found in service account kube-system/default",
      }
  occurred
  In [It] at: /go/src/github.com/stolostron/multicloud-operators-foundation/test/e2e/managedcluster_test.go:164 @ 11/10/23 02:43:19.359
```

zhujian7:
root cause: Starting Kubernetes 1.24, Secrets are not automatically generated when Service Accounts are created.

PR #681, opened 2023-11-10
stop running get ca from service account test for ocp 1.24 or above

zhujian7:
issue reference: https://github.com/stolostron/multicloud-operators-foundation/issues/680


zhujian7:
/hold
we can only merge this after ACM 2.9 is released

PR #695, opened 2023-11-30
chore: remove refs to deprecated io/ioutil

testwill:


xuezhaojun:
@testwill It seems RedHat Trusted App Pipeline can't be triggered by a non-org-member account, would you mind to apply join the org first?

PR #705, opened 2024-01-30
refactor log using clusterproxy and managedserviceaccount addons

zhiweiyin318:
story: https://issues.redhat.com/browse/ACM-9195


pkg/proxyserver/getter/loggetter.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"time"
+
+	utilnet "k8s.io/apimachinery/pkg/util/net"
+	"k8s.io/client-go/dynamic"
+	restclient "k8s.io/client-go/rest"
+)
+
+// ClientConfig is to define the configuration to connect klusterlet
+type ClientConfig struct {
```
zhujian7, line 12: I did not see any place that uses this `ClientConfig`

pkg/proxyserver/rest/log/log.go
```diff
@@ -51,56 +53,35 @@ func (r *LogRest) Connect(ctx context.Context, id string, opts runtime.Object, r
...
+		PodNamespace:  pathArray[1],
+		PodName:       pathArray[2],
+		ContainerName: pathArray[3],
+		Query:         map[string][]string{"tailLines": {"1000"}},
+	}
+	klog.Infof("### handler: %+v", logHandler)
+
+	return &logHandler, nil
+
+	// location.Path = netutil.JoinPreservingTrailingSlash(location.Path, proxyOpts.Path)
```
zhujian7, line 78: delete?

pkg/controllers/managedserviceaccount/log_serviceaccount_controller.go
```diff
@@ -0,0 +1,133 @@
+package managedserviceaccount
```
zhujian7, line 1: why does this controller belong to the ocm controller instead of the ocm proxyserver?

pkg/proxyserver/rest/log/log.go
zhiweiyin318: done

pkg/proxyserver/getter/loggetter.go
zhiweiyin318: deleted

pkg/controllers/managedserviceaccount/log_serviceaccount_controller.go
zhiweiyin318: The log permission rbac is managed by the ocm-controller, so I think it makes sense to create the managed SA in the ocm-controller.

pkg/addon/manifests/chart/templates/log_service_account.yaml
```diff
@@ -0,0 +1,10 @@
+{{- if ne .Values.installMode "Hosted" }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "work-manager.fullname" . }}-log
```
qiujian16, line 5: why do we need this? I thought we use msa?

pkg/klusterlet/clusterinfo/loggingInfo_sync.go
```diff
@@ -2,329 +2,60 @@ package controllers
...
 	}
 
-	endpoint.Hostname = route.Spec.Host
-	return endpoint, nil
-}
-
-func (s *loggingInfoSyncer) getOwnerRef(ctx context.Context, ocp311 bool) (*metav1.OwnerReference, error) {
-	if ocp311 {
-		deploy, err := s.managementClusterClient.ExtensionsV1beta1().Deployments(s.agentNamespace).Get(ctx, s.agentName, metav1.GetOptions{})
+	_, err = s.routeV1Client.RouteV1().Routes(s.agentNamespace).Get(ctx, s.agentName, metav1.GetOptions{})
```
qiujian16, line 50: I am not sure if it is needed. I do not think we need to delete the route. This could be doced and let user do it manually.

deploy/foundation/hub/rbac/clusterrole-foundation.yaml
```diff
@@ -105,3 +105,9 @@ rules:
   - apiGroups: ["addon.open-cluster-management.io"]
     resources: ["addondeploymentconfigs"]
     verbs: ["get", "list", "watch"]
+  - apiGroups: ["config.openshift.io"]
+    resources: ["ingresses"]
```
qiujian16, line 109: why we need this?

pkg/proxyserver/log/logClient.go
```diff
@@ -0,0 +1,117 @@
...
+	logTokenSecret, err := c.KubeClient.CoreV1().Secrets(c.ClusterName).Get(context.Background(),
+		workManagerLogSAName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	userServerHost := "cluster-proxy-user." + c.BaseDomain
+
+	clusterProxyCfg := &rest.Config{
+		Host: fmt.Sprintf("https://%s/%s", userServerHost, c.ClusterName),
```
qiujian16, line 42: I think we should directly use grpc client for cluster-proxy instead of using rest service. You can check how it works in clusteradm.

pkg/proxyserver/controller/controller_test.go
```diff
@@ -84,19 +88,49 @@ func newTestContent(t *testing.T) *testContent {
 
 func (c *testContent) newController() {
 	kubeClient := k8sfake.NewSimpleClientset(c.objects...)
-	c.kubeSharedInformers = kubeinformers.NewSharedInformerFactory(kubeClient, 10*time.Minute)
+	c.proxyServiceConfigmapInformer = kubeinformers.NewSharedInformerFactoryWithOptions(kubeClient,
```
qiujian16, line 91: what is this configmap used for?

PR #707, opened 2024-01-31
Add global placement to select all clusters

zhujian7:


zhiweiyin318:
need add placement rbac to foundation clusterrole. also need update the clusterrole in mce.

zhiweiyin318:
can be merged after add rbac in MCE operator

PR #729, opened 2024-03-25
Adopt install strategy in foundation addon

qiujian16:


deploy/foundation/hub/ocm-controller/clustermanagementaddon.yaml
```diff
@@ -2,10 +2,17 @@ apiVersion: addon.open-cluster-management.io/v1alpha1
...
  name: work-manager
+ annotations:
+   "addon.open-cluster-management.io/lifecycle": "addon-manager"
 spec:
   addOnMeta:
     displayName: work-manager
     description: "work-manager provides action, view and rbac settings"
   supportedConfigs:
   - group: addon.open-cluster-management.io
     resource: addondeploymentconfigs
```
zhujian7, line 13: if users want to configure the install ns by an addondeplymentconfi, they need to add default config here right? will installer change the default value back?
Or users should change ns per cluster by adding addondeploymentconfig on mca?

qiujian16: how about we add a default empty config in the installer

zhujian7: checked in my dev env, add default config will not be reconciled back by installer.

PR #737, opened 2024-04-16
update addon-framework to fix addon config remove issues

haoqing0110:
Fix: https://issues.redhat.com/browse/ACM-10982


qiujian16:
does it need to be cherry-picked?

PR #746, opened 2024-05-08
Fix: no update when after set clusterclaim.

xuezhaojun:


test/e2e/util/hive.go
```diff
@@ -200,5 +200,7 @@ func ClaimCluster(hiveClient hiveclient.Interface, cdName, cdNs, claimName strin
 		return err
 	}
 	cd.Spec.ClusterPoolRef.ClaimName = claimName
-	return nil
+
+	_, err = hiveClient.HiveV1().ClusterDeployments(cdNs).Update(context.TODO(), cd, metav1.UpdateOptions{})
+	return err
```
xuezhaojun, line 205: It seems we attempt to add the clusterclaim to clusterDeployment, so I add the `Update`. @ldpliu

PR #753, opened 2024-05-27
Create addon ns when it is not the default addon ns

qiujian16:


pkg/addon/manifests/chart/templates/namespace.yaml
```diff
@@ -1,12 +1,16 @@
-{{- if eq .Values.installMode "Hosted" }}
+# namespace is klusterlet-{clusterName} in hosted mode, and we need to
+# set specific label/annotation.
+{{- if ne .Release.Namespace "open-cluster-management-agent-addon" }}
 apiVersion: v1
 kind: Namespace
 metadata:
   name: {{ .Release.Namespace }}
   labels:
+  {{- if eq .Values.installMode "Hosted" }}
```
zhujian7, line 9: does this also need label `addon.open-cluster-management.io/namespace` like https://github.com/open-cluster-management-io/managed-serviceaccount/pull/109

qiujian16: this label has been added I think

PR #754, opened 2024-06-04
Add hosted addon annotation to mca

qiujian16:


pkg/controllers/hosteannotation/hosted_annotation_controller.go
```diff
@@ -0,0 +1,143 @@
...
+	addOn := &addonapiv1alpha1.ManagedClusterAddOn{}
+	err := r.client.Get(ctx, types.NamespacedName{Name: addonName, Namespace: cluster.Name}, addOn)
+
+	if errors.IsNotFound(err) {
+		return nil
+	}
+
+	_, hostingClusterName := addon.HostedClusterInfo(addOn, cluster)
+
+	if len(hostingClusterName) == 0 {
```
elgnay, line 131: The add-on is not always deployed in the hosted mode when the klusterlet is in the hosted mode. Instead The add-on is deployed in the hosted mode only when
- the klusterlet is in the hosted mode;
- the ManagedCluster has an annotation `addon.open-cluster-management.io/enable-hosted-mode-addons: "true"`;
- and it supports the hosted mode;

qiujian16: this is checked in HostedClusterInfo func

PR #755, opened 2024-06-05
update ocp version if there is only 1 history item in clusterVersion

zhiweiyin318:
issue : will not get the OCP version to the clusterClaim when the history items in the clusterversion only has 1 item and the status is not Completed.
fix: get the OCP version if the history items in the clusterversion only has 1 item, will not check if the item is Completed.


xuezhaojun:
Hi, @zhiweiyin318 seems https://issues.redhat.com/browse/ACM-12011 is closed,  should we also close or merge the PR?

PR #759, opened 2024-06-19
Change clusterrolebinding namespace by install ns

qiujian16:


pkg/addon/manifests/chart/templates/cluster_role_binding.yaml
```diff
@@ -2,7 +2,11 @@
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
+  {{- if eq .Release.Namespace "open-cluster-management-agent-addon" }}
   name: {{ .Values.org }}:{{ template "work-manager.fullname" . }}
+  {{- else }}
+  name: {{ .Values.org }}:{{ template "work-manager.fullname" . }}:{{ .Release.Namespace }}
```
zhujian7, line 8: Is it possible to exceed the length limit?

qiujian16: the name is 253, while the namespace limit is 64

PR #773, opened 2024-08-28
Upgrade hive apis.

xuezhaojun:
Fixes: https://issues.redhat.com/browse/ACM-13469


pkg/controllers/clusterset/clustersetmapper/clusterset_controller_test.go
```diff
@@ -311,11 +312,16 @@ func TestReconcile(t *testing.T) {
 			r := newTestReconciler(test.existingObjs, test.existingRoleObjs, test.globalSetMapper, test.clusterSetClusterMapper, test.clusterSetNamespaceMapper)
 			r.Reconcile(ctx, test.req)
 
-			clusterroleName := utils.GenerateClustersetClusterroleName(ManagedClusterSetName, "admin")
-			clusterrole, _ := r.kubeClient.RbacV1().ClusterRoles().Get(context.TODO(), clusterroleName, metav1.GetOptions{})
-
 			var clusterroleExist bool
-			if clusterrole != nil {
+			clusterroleName := utils.GenerateClustersetClusterroleName(ManagedClusterSetName, "admin")
+			_, err := r.kubeClient.RbacV1().ClusterRoles().Get(context.TODO(), clusterroleName, metav1.GetOptions{})
```
xuezhaojun, line 317: @zhujian7 There is a behavior change:
* before when “Not Found”,  the result is nil
* after when “Not Found”, the result is not nil

xuezhaojun: So I changed to use err to determine whether clusterrole exists.

PR #779, opened 2024-09-19
fix cannot override the addon values using customized variable

zhiweiyin318:
issue: https://issues.redhat.com/browse/ACM-14178

### DIFF
--- a/pkg/addon/addon.go
+++ b/pkg/addon/addon.go
@@ -48,6 +48,11 @@ var ChartFS embed.FS
 
 const ChartDir = "manifests/chart"
 
+const (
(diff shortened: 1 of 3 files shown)


pkg/addon/addon.go
```diff
@@ -48,6 +48,11 @@ var ChartFS embed.FS
 
 const ChartDir = "manifests/chart"
 
+const (
+	strTrue  string = "true"
+	StrFalse string = "false"
```
qiujian16, line 53: why capital S?

zhiweiyin318: typo.
fixed.

PR #852, opened 2025-02-07
upgrade: update Go version to 1.23

xuezhaojun:


zhiweiyin318:
@xuezhaojun do we need to change the prow job too ?

xuezhaojun:
> @xuezhaojun do we need to change the prow job too ?

Yes, I can do that

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.ci.openshift.org/stolostron/builder:go1.22-linux AS builder
+FROM registry.ci.openshift.org/stolostron/builder:go1.23-linux AS builder
```
zhujian7, line 1: follow-up:
- [ ] update the gitlab Dockerfile: https://gitlab.cee.redhat.com/mce-cpaas-midstream/multicloud-manager/-/blob/multicluster-engine-2.8-rhel-9/distgit/containers/multicloud-manager/Dockerfile.cached?ref_type=heads#L1

PR #980, opened 2025-05-13
Change `any` to * in list kubevirt projects api

jlpadilla:
Change `any` to `*` in the kubevirtprojects api response to avoid potential issues if a project is named "any" in the managed cluster.

/assign @skeeey

### DIFF
--- a/docs/clusterview/clusterview.md
+++ b/docs/clusterview/clusterview.md
@@ -289,7 +289,7 @@ The `kubevirtprojects.clusterview.open-cluster-management.io` API is only used t
     cluster2   kubevirt-workspace-2
     ```
 
-**Note**: If using ClusterRoleBinding in ClusterPermission to bind a kubevirt ClusterRole, this API will return `any` in project field, for example:
+**Note**: If using ClusterRoleBinding in ClusterPermission to bind a kubevirt ClusterRole, this API will return `*` in project field, for example:
 
 ```yaml
 apiVersion: rbac.open-cluster-management.io/v1alpha1
@@ -329,7 +329,7 @@ spec:
 kubectl get kubevirtprojects.clusterview.open-cluster-management.io
 
 CLUSTER    PROJECT
-cluster1   any
-cluster2   any
+cluster1   *
+cluster2   *
 ```
 
--- a/pkg/proxyserver/rest/project/list.go
+++ b/pkg/proxyserver/rest/project/list.go
@@ -44,7 +44,7 @@ func listProjects(namespace, name string, obj runtime.Object, userInfo user.Info
 			}
 
 			if isKubeVirtRole(roleName) {
-				projects = append(projects, projectView{name: "any", cluster: namespace})
+				projects = append(projects, projectView{name: "*", cluster: namespace})
 				return projects
 			}
 		}


skeeey:
@jlpadilla  because the value of project column comes from the ocp project name, and the `*` is a special character, it is not allowed for project name, if we use `*`,  it may cause other problems in future, e.g. if we provide a get api for kubevirtprojects, it will not work

Can we use some other name for this, e.g. `all-projects`, `all-managedcluster-projects` or `all-kubevirt-projects`, but I think the name should follow the k8s name specification,  https://pwittrock.github.io/docs/concepts/overview/working-with-objects/names/ , what's your opinion？

jlpadilla:
@skeeey The problem with following the k8s name spec is that the user could have a real project with the same name. We don't have any way to know if it's a real project or not. A malicious user could exploit it to break our RBAC logic.

By using a special character like `*` we can be sure that there isn't a real project with that name.

skeeey:
I think we may be able to use the clustername+projectname as the KubevirProjectView name in kubevirtprojects api, and  if the KubevirProjectView represents all namespaces for one cluster, its name will be clustername, e.g.

```
- apiVersion: clusterview.open-cluster-management.io/v1
  kind: Project
  metadata:
    labels:
      cluster: cluster1
    name: cluster1 <-----  this means the all projects of cluster1
- apiVersion: clusterview.open-cluster-management.io/v1
  kind: Project
  metadata:
    labels:
      cluster: cluster2
    name: cluster2-any <-----  this means the project any of cluster2
```

and we enhance the kubevirtprojects list view logic, if one projectview is for cluster all namespaces, we can make it to show
```
CLUSTER    PROJECT
cluster1       *
```

here is some draft code https://github.com/stolostron/multicloud-operators-foundation/pull/982/files

@jlpadilla  WDYT

/cc @qiujian16

jlpadilla:
@skeeey I think this works.
If I had a real project named `cluster2-any` it would become `cluster2-cluster2-any` which is unique.

But I wonder if the KubevirProjectView name length could cause problems. When you combine cluster+namespace the value could be more than the 63 characters allowed for a resource name.

skeeey:
> But I wonder if the KubevirProjectView name length could cause problems. When you combine cluster+namespace the value could be more than the 63 characters allowed for a resource name.

yes, this will be a problem, how about we use an uuid as the KubevirProjectView name and use a label `project` to reparents the project name

- if the KubevirProjectView reparents to one certain project for a cluster
```
- apiVersion: clusterview.open-cluster-management.io/v1
  kind: Project
  metadata:
    labels:
      cluster: cluster1
      project: test
    name: 6007d213-edcd-4ffd-8c5b-c7b570f92648
```

- if the KubevirProjectView reparents to all projects for a cluster, we will use `all_projects` as the label value (the `all_projects` cannot be used as name for kube resource, this will ensure there are no namespaces named `all_projects`)
```
apiVersion: clusterview.open-cluster-management.io/v1
  kind: Project
  metadata:
    labels:
      cluster: cluster2
      project: all_projects
    name: 6007d213-edcd-4ffd-8c5b-c7b570f92649
```

jlpadilla:
The solution with uuid and labels should work. Thanks @skeeey !
I'll close this in favor of PR #982

PR #1017, opened 2025-06-12
fix doc issue

skeeey:


zhiweiyin318:
/override "Red Hat Konflux / multicloud-manager-mce-29-on-pull-request"

PR #1227, opened 2026-03-23
ACM-31402: use OpenAPI model name package to avoid JSON Pointer encoding issue in schema references

zhujian7:
## Summary

- Add `+k8s:openapi-model-package` annotations to vendor API packages (`open-cluster-management.io/api`, `github.com/stolostron/cluster-lifecycle-api`) so that `OpenAPIModelName()` methods are generated for each type, returning dot-format names (e.g. `io.open-cluster-management.api.cluster.v1.ManagedCluster`).
- Update `zz_generated.openapi.go` to use these methods instead of hardcoded Go package path strings (e.g. `"open-cluster-management.io/api/cluster/v1.ManagedCluster"`).
- Replace local path replace directives with published GitHub forks that contain the upstream `OpenAPIModelName()` changes.

## Root cause

Old Go package path definition names contain `/` characters. When kube-openapi builds `$ref` values it calls `EscapeJsonPointer()` which encodes `/` as `~1` (RFC 6901). However `document.go` does **not** decode `~1` back to `/` when looking up models in the definitions map, causing:

```
SchemaError(open-cluster-management.io/api/cluster/v1.ManagedCluster.spec):
  unknown model in reference: "open-cluster-management.io~1api~1cluster~1v1.ManagedClusterSpec"
```

The new dot-only format contains no `/`, so `EscapeJsonPointer` is a no-op and all `$ref` lookups succeed.


chdeshpa-hue:
This fix needs to be cherry-picked to `backplane-2.11` for MCE 2.11.3.

MCE 2.11.2 shipped without this fix and is actively broken in production — the `ocm-proxyserver` publishes corrupted OpenAPI `$ref` references (`~1` escaped slashes vs unescaped definition keys), breaking ArgoCD cluster cache sync, `oc explain`, and all kube-openapi consumers on the hub cluster. Multiple customers affected, no workaround exists.

The fix has been on `main` since March 2026 but was not backported to the 2.11 release branch. Requesting cherry-pick to `backplane-2.11` as a Sev1/urgent z-stream fix.

**Error signature on affected clusters:**
```
SchemaError(github.com/stolostron/cluster-lifecycle-api/clusterview/v1alpha1.UserPermission.status): 
unknown model in reference: 
"github.com~1stolostron~1cluster-lifecycle-api~1clusterview~1v1alpha1.UserPermissionStatus"
```

PR #1284, opened 2026-05-14
Fix e2e so we can pass in branch name in CI

tesshuflower:
- We previously fixed the CI e2e jobs (https://github.com/openshift/release/pull/79263) to pass in either the correct image:tag, or for other images such as these ones in deploy/managedcluster/addons/install.sh (which already specify the stolostron version of images), we set OCM_BRANCH.

- However cluster-proxy is using hardcoded "main" in the e2e install script rather than the provided OCM_BRANCH - which means e2e tests are always picking the latest cluster-proxy image (with `main` tag).


dhaiducek (approved):
LGTM! Thanks for catching that!

tesshuflower:
Looks like we don't need the cherry-pick for `backplane-2.6` as it's hardcoding the correct branch:

https://github.com/stolostron/multicloud-operators-foundation/blob/9fd2e6af3c68b6f90464238ac7e0f9a318986fce/deploy/managedcluster/addons/install.sh#L64-L65

Same things for backplane-2.7-2.11

We did need this cherry-picked into release-2.17 - so will approve that one.

PR #1291, opened 2026-05-21
ACM-34295: Fix stale nodeList in ManagedClusterInfo after node deletion

asadawar:
## Summary

- Register a `DeleteFunc` event handler on the node informer to trigger immediate reconciliation when nodes are deleted, reducing stale nodeList latency from up to 10 minutes to ~1 second
- Replace `wait.JitterUntilWithContext(context.TODO(), ...)` with a select-loop that properly respects context cancellation, fixing a goroutine leak on agent restart
- Add test case for the node deletion scenario (3 stale nodes in ManagedClusterInfo, 1 in informer cache → verify cleanup)

## Root Cause

The node collector at `pkg/klusterlet/nodecollector/nodecollecter.go` polled the informer cache every 30s but did not register any event handlers. When node deletions occurred (e.g., KubeVirt nodepool removal in HCP clusters) and the informer's watch missed the delete events, stale entries persisted in `ManagedClusterInfo.status.nodeList` until the next full informer relist (10 minutes).

A secondary bug: `Start()` used `context.TODO()` instead of the passed `ctx`, so the reconciliation goroutine never stopped on shutdown — causing goroutine leaks and potential stale data on agent restart.

## Test plan

- [x] `go test ./pkg/klusterlet/nodecollector/...` — all 12 tests pass (including new "nodes removed from cache")
- [x] `go vet ./pkg/klusterlet/nodecollector/...` — clean
- [x] `gofmt` — clean
- [x] Runtime-verified on HCP cluster with KubeVirt nodepools (details below)

## Runtime-verified test

**Environment:**
- Hub: OCP 4.19.7, ACM 2.16.1, MCE 2.11.1
- Hosted cluster: OCP 4.20.6, KubeVirt provider
- Patched image: `quay.io/asadawar/multicloud-manager:acm-34295`

**Procedure:**
1. Built patched `multicloud-manager` binary with the fix and deployed as standalone agent on the hosted cluster
2. Scaled KubeVirt nodepool from 2 → 1 replica
3. Node deleted ~3 minutes after scale-down
4. Checked `ManagedClusterInfo.status.nodeList` on hub — **updated to 1 node within ~15 seconds of deletion**
5. Scaled nodepool back to 2 replicas, confirmed both nodes healthy

**Before fix:** stale nodes persisted in `ManagedClusterInfo.status.nodeList` for up to 10 minutes after deletion, causing Fleet Management UI to show deleted nodes as "Ready".

**After fix:** `ManagedClusterInfo.status.nodeList` updated within seconds of node deletion via the new `DeleteFunc` event handler.

Fixes: https://redhat.atlassian.net/browse/ACM-34295

### DIFF
--- a/pkg/klusterlet/nodecollector/nodecollecter.go
+++ b/pkg/klusterlet/nodecollector/nodecollecter.go
@@ -70,6 +70,7 @@ type queryResult struct {
 type resourceCollector struct {
 	nodeLister         corev1lister.NodeLister
 	nodeSynced         cache.InformerSynced
+	reconcileCh        chan struct{}
 	kubeClient         kubernetes.Interface
 	client             client.Client
 	clusterName        string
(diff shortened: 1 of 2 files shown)


dhaiducek:
/ok-to-test

Thanks for the update @asadawar! I think this update makes sense.

dhaiducek (approved):
/hold for other reviews

PR #1323, opened 2026-07-08
update to golang1.26

tesshuflower:
Also removes the old incorrect cpe value from Dockerfile.rhap - the cpe label will be inserted automatically by the common pipeline.


dhaiducek (approved):
LGTM!

PR #1353, opened 2026-08-27
Filter node heartbeats in ClusterClaim reconciler

smunaval:
## Summary
- Fixes ACM-42809: klusterlet-addon-workmgr reconciled ClusterClaims on every kubelet node heartbeat because the Node informer ignored predicates and each reconcile live-GETed every ClusterClaim.
- Ignore heartbeat-only Node updates. Reconcile only when unschedulable, taints, capacity, or Ready/NetworkUnavailable change.
- Read existing ClusterClaims from the informer cache instead of live API GETs, and DeepCopy before update.

Lab note: on ACM 2.17 / 6 nodes, ClusterClaim writes were already skipped when spec was unchanged. The missing predicates and live GETs are still a performance bug on large managed clusters (customer case 04508842).

## Test plan
- [x] go test ./pkg/klusterlet/objectsource ./pkg/klusterlet/clusterclaim
- [ ] Verify schedulable.open-cluster-management.io still updates when a node is cordoned/uncordoned
- [ ] On a managed cluster, confirm node heartbeats no longer enqueue ClusterClaim reconciles / live GETs

### DIFF
--- a/pkg/klusterlet/clusterclaim/clusterclaim_controller.go
+++ b/pkg/klusterlet/clusterclaim/clusterclaim_controller.go
@@ -122,7 +122,7 @@ func (r *clusterClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 }
 
 func (r *clusterClaimReconciler) syncControlPlaneCreatedClaims(ctx context.Context) error {
-	return syncClaims(ctx, r.clusterClient,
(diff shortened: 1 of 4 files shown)


dhaiducek:
/retest


Retrying due to Konflux infra error:
```
Error allocating host: failed to provision host
```

